PR #6, opened 2014-03-12
Renames tgstation.dme to NTstation13.dme in order to avoid the safety ch...

PolymorphBlue:
...eck when testing with the default folder name

Changes most of the branding along with it at the same time just because
Adds myself to the coder's admins.txt for convenience.


PolymorphBlue:
Just going to commit this, if anyone has an issue please comment

KazeEspada:
One issue is that it breaks the maintainer can't merge their own pull rule. Usually a pull like this can just be pulled by the next available maintainer.

PR #386, opened 2014-03-18
Changelog for borg renaming.

Neerti:
Self explanatory.

Not sure why the high remove/add count but it should check out.


StarToad:
That it does, but i'm still unsure why. Ah well. If it fucks the Changelog up irrevocably we can always revert this. :V

vista-:
use the goddamn tools that are provided
you know, the changelog merger tool

Neerti:
I used the addtochangelog tool.

PR #393, opened 2014-03-27
Detective Work (trace evidence) [Ready]

Petethegoat:
Added gun trace evidence.
Adds evidence on the shooter, the victim (including hit objects), the gun, and the casings.
Adds appropriate descriptions for most projectiles. Laser guns don't leave traces on the shooter.

Doubled and then some the chance for fibers to be placed on objects and clothing.

Adds fibers from the victim to weapons.

Unrelated fixes to make eguns and coins show up in the map editor.


Petethegoat:
Oh, this also slightly adjusts the sabre sprite, (R&D SMG) to make the foregrip less fat. Nothing major.

Petethegoat:
It is supposed to be two underscores, yes.

CollenN:
I like this change.

Perhaps something that might be neat would be ballistic markings on bullet casings and guns, allowing detectives to match bullet casings with the gun that made them?

Petethegoat:
That's definitely something I've thought about, although I'm not sure how much use it'd get (how often are there going to be multiple possible guns for you to compare with?)
But I'll keep it in mind.

Atskadan:
Inspecting bullets already tells you what caliber they're in, and you can easily narrow down the list with that. Outside of summon guns there's only a few guns that fire bullets. I think scanning a person who has been shot or the clothes of someone who has been shot should show a bullet hole of a certain size for each caliber.

Petethegoat:
I'm happy with this. Merge at will.

Atskadan:
We can’t automatically merge this pull request.

Petethegoat:
Fixed.

PR #394, opened 2014-03-27
Removes the genetics powers virus symptom. [Don't Merge]

Petethegoat:
Removes the genetics powers virus symptom.
The disability one remains, for now.
Plus a bonus typo fix.

Wait for more feedback in http://www.ss13.eu/phpbb/viewtopic.php?f=42&t=4075 before merging.

### DIFF
--- a/code/datums/diseases/advance/symptoms/genetics.dm
+++ b/code/datums/diseases/advance/symptoms/genetics.dm
@@ -6,7 +6,7 @@ DNA Saboteur
 	Very noticable.
 	Lowers resistance tremendously.
 	No changes to stage speed.
-	Decreases transmittablity temrendously.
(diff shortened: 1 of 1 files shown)


GunHog:
As I said on the thread, I do not support this one. Removing hulk is a more acceptable nerf in my opinion.

Petethegoat:
I'll probably change the implementation, depending on the feedback. This almost certainly isn't the final PR.

KazeEspada:
Although I haven played in a while, wouldn't it be better to make it even harder to get? or even just removing it from the possibility of appearing without admin help.  I don't like the idea of removing content from virology.

Petethegoat:
Given that the thread has died I think I'll close this for now.

PR #402, opened 2014-04-03
Adds the Ancient Hardsuit.

MrSnapwalk:
Sprites scavenged from the old /tg/station forums. Adds a hardsuit from the year 2301. It slows down the user due to it's bulk, but provides more protection. Designed to be found as loot in space. Please merge this rather quickly, as I'm trying to use it in a space explorable in my new map.


Atskadan:
I don't see why an older version of hardsuit would be better than the new ones?

MrSnapwalk:
Well, it's not really better. It's quite bulky due to the heavier plating, and therefore slows you down quite a bit. But, seeing as the plating is thicker, it protects you more. Kind of a trade-off, and plus: this thing is rare, and there are only supposed to be one or two of them on the map somewhere in deep-space. It's kind of like the voidsuit in a way.

Atskadan:
Ah, sure. Well, fluff doesn't matter anyway. What's the sprite?

MrSnapwalk:
Found the mob-sprites on the old /tg/station forums, and I made the object and in-hand sprites.

Atskadan:
Don't see why not.

PR #405, opened 2014-04-07
Updates the kitchen to the design I made earlier

Atskadan:
I think this design is much nicer and offers more room for the chef to work.


Atskadan:
oh i forgot to use the merge tool
oh well

PR #406, opened 2014-04-07
Puts in a plethora of pleasant peanut paraphernalias [DON'T MERGE]

Atskadan:
Peanuts and peanut accessories.

For Secret


Atskadan:
maybe later

Petethegoat:
yeah peanut allergies would be hilarious. actual uses for medical records are always good too!

sitp:
thank you miggles wiggles <3 shake girl will commence

PR #410, opened 2014-04-08
[MAP] Metastation Kitchen + Merge tools

Atskadan:
Updates the Metastation map's kitchen to include the new chef machines (and remodels the kitchen itself a bit). The map merge tools now work with Metastation instead of Box.


Atskadan:
apologies again for not using the merge tool, but my map corrupted and i had to redownload a fresh copy

sitp:
that enzyme bottle is covering up the way to the delicious deep fryer, does it work even with a table there?

Yakk:
All that empty space where tables for ingredients for mircowave dishes could go.

e: oh wait the back room has tables now those have a use I guess my tism is sated somewhat

Atskadan:
@sitp Yes, I checked that it works.
@Yakk what do you mean? I specifically added those extra 3 tables so you could put stuff there.

Yakk:
I mean box's middle table is 2x4 and this one is 2x3
You can move junk around on it easier to make space and you have space for ingredients like flour, eggs, and cheese wedges so you don't have to move more than 3 tiles when you start microwaving complicated recipes.

Iamgoofball:
That deep fryer in the corner is inaccessable

Atskadan:
I tested that same corner with a cereal machine, so I assume it would be the same with a deep fryer. Are they different in how they handle items or something? Because I could place it in the cereal machine.

Metacide:
:+1: - this new layout looks solid.

Are the new machines constructible / deconstrictible, or can the be unanchored for movement or anything like that? Just out of interest, as the tabling can always be remodelled easily according to the whims of the chef with the gear in the kitchen backroom closet.

Atskadan:
I think that if non-edible items are ever made unfryable, then unwrenching the machines should also be included. As of yet, neither of those things are true, yet, though.

PR #412, opened 2014-04-09
[EMERGENCY] Fixes undershirts so that they do not look like they are made of cardboard boxes.

MrSnapwalk:
SO BOXY, OH GOD. (Couldn't do anything about the shirts I took from Bay, though. They'll just look boxy, I guess.)


RemieRichards:
> Couldn't do anything about the shirts I took from Bay

Right click the side images in that little 4 direction view, export as Png, Fix it, Stick it back in.

MrSnapwalk:
What I mean is that I literally can't make any modifications that make it look better. If I remove the offending pixel, it looks absolutely crap due to the skin showing up underneath.

RemieRichards:
but it does the same with your Tank tops too, does it not?
I tested the pixel removal on your Sec one and a pixel does show through.

MrSnapwalk:
Yes, but with tank tops it actually makes sense and does not look like crap (as opposed to the boxes we accidentally have now).

RemieRichards:
oh fuck, Sorry my head's still a mess since my news, you said "Shirts" didn't you? Sorry, I got confused.

PR #418, opened 2014-04-10
Heads may use ID Modification Consoles to manage their department

ghost:
Poll: http://www.ss13.eu/phpbb/viewtopic.php?f=42&t=4307

The primary goal is to allow department heads to demote their own staff. The secondary goal is to let them give others access to their department.

screenshot 2014-04-11 20 51 36
-Heads may modify ID access for their department using ID computers
-Heads may demote subordinate jobs
-Heads may not modify a card's name or title, except when Demoting an ID


RemieRichards:
Actually, Ikky took the easiest route there, Byond allows both declarations, however if a majority of a file is written one way, I've had byond happily REFUSE valid declarations, just because it isn't "Consistent" to the rest of the file.

Ikky would have to change every Deceleration in the file to full typepaths (Best Result)
or simply we accept this file being odd (Like a bunch of the older files, Ok Result)

Atskadan:
i was making a joke

ghost:
I'm getting a bizarre issue of the computer only fetching a partial list of the RD's access. If anyone can help me figure out why I'd really appreciate it.

Note how tech storage and AI upload is missing

ghost:
Revised Department Management Functions

-Heads may only edit access to their own departments
-Demotion button will check if the job is a subordinate of the head instead of checking if the person has any access left

ghost:
Good to go!

PR #419, opened 2014-04-10
Stun time rebalancing

ghost:
EDIT 4/16:

The movement speed changes have been removed and move code will remain unaffected as it is far too great of an issue to tackle at the moment. All this PR does in its current state is cut 20% off of most stuns, which should balance the game a little better.

These stun time changes are now merge-ready after passing forum polling.


ghost:
I did, but it's rather hard to tell.. If it hasn't done anything other than automatically go -1 for the shoe speedup then I'll do additional code spelunking and try to sort it out.

ghost:
Base run speed is now slightly higher than barefoot used to be. Shoes do not speed you up or slow you down. I'm mad as fuck this isn't pull #420.

Intigracy:
The define does decimals.
So does the configuration option for movespeed delay.

Something I have been saying repeatedly.

ghost:
Well, I removed shoe speedup entirely because a lot of people like that in comparison to the old way shoes were handled. Base movement speed is faster, though.

Intigracy:
You could make hardsuits not have suit slowdown in low pressure environments.

ghost:
i could, i'm looking at ergo's code for that actually, the way he makes the miner shotgun do more damage in low pressure.

ghost:
so yeah, going to leave this PR alone for now, but all it does in its current state is lower stun times a little bit from the ridiculousness that is 10 cycle/second stuns.

PR #424, opened 2014-04-11
Adds telnyashka shirts and tank tops

ghost:
-also fixes a minor issue with the treehugger tank top's name


ghost:
clipboard01
clipboard02

russia intensifies

PR #429, opened 2014-04-12
Borging removes all Antag Status, Borg Memory Config setting [Good to Go]

RemieRichards:
Idea shamelessly nicked from Pete's TODO list (I figure I'd take something off your load)

Adds remove_antag(specified_antag) which removes all references to an antag type inside a mind
Update mind.dm's href "clear" references to use remove_antag()
remove_antag() is designed to be run on a mind
adds remove_all_antag() which calls remove_antag() on all antag types possible
wipe_memory() is just a quick shortcut for memory = null

These are MIND PROCS, Don't run them on mobs, run them on the mob's mind.

Config setting so Borg's no longer retain memories after borging, This was intended functionality, But rather than do it somebody put a message telling borg's to ignore their memories, God I love the Ancient coders.

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -99,6 +99,7 @@
 	var/revival_brain_life = -1
 
 	var/rename_cyborg = 1
+	var/borg_remembers = 0
 
 	//Used for modifying movement speed for mobs.
 	//Unversal modifiers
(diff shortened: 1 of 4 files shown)


Petethegoat:
Maybe change the text to be more explicit, like "You forget everything about your past life!" or something. But otherwise, this looks good, thanks for doing it. :+1: 

`//Wipes memories - RR` is a pretty superfluous comment though!

RemieRichards:
@Petethegoat I'll run a pass on commenting more/less where necessary, Just running some tests.
And yeah, your message is better, Stealing.

Petethegoat:
Cool. Don't forget to make a poll!

RemieRichards:
Poll is largely in favour, 12 yes, 6 no, 2 abstentions.

PR #431, opened 2014-04-12
[Port] Emergency Shuttle Signal Tracing

ghost:
Poll: http://www.ss13.eu/phpbb/viewtopic.php?f=48&t=4342

Intended to cut down on shuttle ping-ponging caused by stealth comms consoles.

-After the second recall, centcom will attempt to trace the origin of emergency shuttle calls and recalls
-The results can be viewed on any authenticated communications console.
-Tracing has a 40% failure chance to further reduce the risk of metagaming the game mode

Ported from TGS


Atskadan:
This feature didn't really require a poll, so I'll merge it now.

PR #432, opened 2014-04-12
[Port] Updated the godawful Airlock Electronics interface

ghost:
Also added the ability to set if the door should require ALL listed accesses, or just ONE.

screenshot 2014-04-06 15 00 55

Ported from TGS


Atskadan:
holy shit yes please

PR #440, opened 2014-04-13
Fixes a critical exploit with the autolathe.

Walter0o:
While browsing through my server's code looking for possible exploits to fix, i noticed the following :

The autolathe can be used to duplicate any and all objs.

The faulty code accepts any refID from the usr without a safetycheck to see if the requested obj is in the autolathe_recipes list.

This works "only" on objs because it will trigger a runtime error if the object has no material vars.

The default buildcost values for obj are zero, so it always goes through the materials-check, but it would not be sufficient to plug this exploit at this point.

The trivial fix is to have a check to see if the given refID is in the autolathe_recipes list, although a datum-based construction method would probably be more robust.

As basically identical autolathe code appears to be used in Baycode , /tg/, /vg/, Para, and all other builds i could look at, i assumed this exploit has been undetected since Goon.

And indeed, the faulty code is present in Gooncode rev4407 and has been ever since.

### DIFF
--- a/code/game/machinery/autolathe.dm
+++ b/code/game/machinery/autolathe.dm
@@ -199,9 +199,42 @@ var/global/list/autolathe_recipes_hidden = list( \
 		if(href_list["make"])
 			var/coeff = 2 ** prod_coeff
 			var/turf/T = get_step(src.loc, get_dir(src,usr))
-			var/obj/template = locate(href_list["make"])
+				
+			// critical exploit fix start -walter0o
+			var/obj/item/template = null
+			var/attempting_to_build = locate(href_list["make"])
+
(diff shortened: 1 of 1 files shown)


Walter0o:
Also reported to Bay, the other /tg/ team, /vg/ and Paradise.
https://github.com/Baystation12/Baystation12/pull/4750
https://github.com/tgstation/-tg-station/pull/3416
https://github.com/d3athrow/vgstation13/pull/626
https://github.com/ZomgPonies/Paradise/pull/179

theOperand:
I think this can be closed, NT pulls from /tg/ periodically.

Atskadan:
Ports like this are always fine, though.

PR #441, opened 2014-04-15
Prevent generating culture bottles with no cooldown

UncleSporky:
I'm new to this so feel free to reject this request if I've screwed something up.

When using the PanD.E.M.I.C 2000 to create culture bottles, it pops up a dialog box asking you to name your culture.  If you ignore this dialog, you can go back to the other window and continue to click "create culture bottle" to keep making them with no cooldown.  The cooldown only kicks in after confirming the dialog box.  Moving the cooldown to before the dialog box fixes the issue.

Again, I'm new here so do with it what you will.  It's as much a test for the process of using git as anything else.  I'll make a post on the forums about this.


Atskadan:
:+1: I'll merge this in a while.

PR #442, opened 2014-04-15
Fixes Inverted Hair Shading

ghost:
When they were first converted to a white base they simply inverted it, making everything backwards

I did not fiddle with the contrast this time and kept it strictly to fixing the inverted highlights, so the lines shouldn't look anywhere near as severe as it may on TG code.

I tested with as many hairstyles as I could but feel free to scream at me this still somehow messes something up, @PinkyGatchabee and I'll correct it by hand.


ghost:
Okay well feel free to ping me with @ikarrus if anything comes up from this

PR #444, opened 2014-04-15
Renames database schema file to generic, branch independent name.

liambaloh:
Renames 'tgstation_schema.sql' to 'database_schema.sql' so that the naming makes sense on both NTStation as well as any branching communities.


Atskadan:
ok sure

PR #446, opened 2014-04-15
Fixes issue with wiki pages requiring cookies even when viewed ingame

liambaloh:
Please note that, legally speaking, this is not a great solution to the problem and may need to be reverted on my end at some point or another. But for right now, it will do.


liambaloh:
Closed due to apostrophes, reopened in pull request #447

PR #447, opened 2014-04-15
Fixes issue with wiki pages requiring cookies even when viewed ingame

liambaloh:
Please note that, legally speaking, this is not a great solution to the problem and may need to be reverted on my end at some point or another. But for right now, it will do.

Also replaces the ’ symbol with a '.


Atskadan:
tell me when this is super-definitely-for-real-this-time ready

liambaloh:
Should be ready. I'm doing these edits directly on github tho, so I'm not compiling them after every edit. But if Travis says it's fine, it's fine.

PR #457, opened 2014-04-17
Readds large crossbow to the protolathe

ghost:
I didn't want this code to go unused, so this readds the large crossbow to the protolathe with:

-New sprites
-New descriptions
-New prices

And it is fairly simple to distinguish the two, so I do not believe there will be "HOLY SHIT TRAITOR LYNCHMOB GO" if someone happens to build one of these with recovered syndie tech.


ChangelingRain:
flat-out double the cost, it's a 5-hit crit and stun through armor, as it's the largebow

ghost:
This should be better to bring it in line with the mini crossbow.

drynwyn:
Note that this can be built without recovering any syndie tech if someone deconstructs a pair of energy relays.

Atskadan:
Yeah, the illegal requirement should be higher.

ghost:
I wasn't aware of the energy relay thing, I'll bump it to 5. :aries:

drynwyn:
I would say just 4. At 5 it requires 2 traitor items deconstructed to produce.

ghost:
you can decon a parapen or a mini ebow to get 5 unless the wiki is outdated, which is the intended reverse engineering process here

ghost:
after testing it, parapen gives 4, so here you go.

Atskadan:
it says merge with caution but i dont see any reason for that 
oh now its gone
well this is pretty good, it still needs tator items to be getted

PR #459, opened 2014-04-18
[Admin] Added logging for changeling hivemind chat

ghost:
Also added their Changeling names to the Antag panel so admins don't have to go digging through vars to find out who was OOC in ICing.


Atskadan:
ok

PR #469, opened 2014-04-19
Fixed Bible Snowflake

RemieRichards:
Who put bible vars in the Gameticker?
Why would you put 1 item's references in the GAMETICKER?
I almost felt physically Ill when I found them.

These vars had one use, Making Library spawned bibles match the current bible, That reeks of Snowflake to me.
# NEW:

Library spawned Bibles randomise their iconstate, itemstate, name and deity name.
This does mean you will get mismatching names, deity names and icons but I think that's ok because you'll end up with some more "unique" religions and bibles from that.

Technically, this means you can no longer get your old bible back if it was destroyed, but then again, No other job unique items have any way to get them back, so screw you snowflake, Protect your bible better!

Moved the chaplain's "Global" bible lists to be real global vars, y'know, So you can actually access them globally! (This was required for the Library Bible Randomization)

Added a deity name list, Apparently we never had one and it's again needed for the Random bibles.
(I just stole a bunch from Games/Fiction/Myhead/Goofball's head)

### DIFF
--- a/code/_globalvars/misc.dm
+++ b/code/_globalvars/misc.dm
@@ -26,5 +26,11 @@ var/global/list/biblestates =		list("bible", "koran", "scrapbook", "burning", "h
 //Bible itemstates
 var/global/list/bibleitemstates =	list("bible", "koran", "scrapbook", "bible", "bible", "bible", "syringe_kit", "syringe_kit", "syringe_kit", "syringe_kit", "syringe_kit", "kingyellow", "ithaqua", "scientology", "melted", "necronomicon")
 
-//Bible deity names (Mainly stolen from 'famous' works of fiction)
-var/global/list/bibledeitynames = 	list("Space Jesus","Space Satan","Cthulu","Armok","Honk Mother","Gia","Beelzebub","Ymir","Zeus","Yog-Sothoth","Talos","Cuban Pete","Chaos","Slaanesh","Death","Drilldoizer","Astaruas","Bringer of Deep fryer","Gwyn","Flying Spaghetti Monster","The Force")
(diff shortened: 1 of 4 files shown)


ghost:
:+1:

code/_globalvars/misc.dm
```diff
@@ -12,3 +12,25 @@ var/timezoneOffset = 0 // The difference betwen midnight (of the host computer)
...
+
+//Bible iconstates
+var/global/list/biblestates =		list("bible", "koran", "scrapbook", "burning", "honk1", "honk2", "creeper", "white", "holylight", "atheist", "tome", "kingyellow", "ithaqua", "scientology", "melted", "necronomicon")
+
+//Bible itemstates
+var/global/list/bibleitemstates =	list("bible", "koran", "scrapbook", "bible", "bible", "bible", "syringe_kit", "syringe_kit", "syringe_kit", "syringe_kit", "syringe_kit", "kingyellow", "ithaqua", "scientology", "melted", "necronomicon")
+
+//Bible deity names, Stolen from Fiction and gathered from the Players
+var/global/list/bibledeitynames = 	list("Space Jesus","Space Satan","Cthulu","Armok","Honk Mother","Gia","Beelzebub","Ymir","Zeus","Yog-Sothoth","Talos","Cuban Pete", \
+"Chaos","Slaanesh","Death","Drilldoizer","Astaruas","Bringer of Deep fryer","Gwyn","Flying Spaghetti Monster","The Force","Great Corgi","Mr Rogers","Old man Henderson", \
```
Atskadan, line 31: >old man henderson
i love it

PR #470, opened 2014-04-20
[DO NOT MERGE] Storage Coats

Atskadan:
Beginning work on porting Bay12's coats with storage. Changes affected coats from /obj/item/clothing/suit/(name) to /obj/item/clothing/suit/storage/(name)
So it changes a lot of maps

Needs extensive testing and the port is not finished yet


RemieRichards:
Keep me informed on Progress with this, I'll help where needed.

Nnelg:
I was working on a new inventory system that would allow you to arbitrarily add and remove slots in a mob's inventory, such as when certain items are equipped (for instance, pockets and the ID slot of a jumpsuit). Using it may be better than trying to code an new type of container that isn't a child of /obj/item/weapon/storage.

I don't know git; is there any way to merge a branch with one in a parallel fork? (I'd like a copy of this code to work off of.)

Atskadan:
So if you have a coat with storage on, it could add 2 pocket slots to your HUD?
I think that could work, but it might lead to cluttered UI. But it does sound like a better idea than mine.

To copy my changes you could try pulling the changes from the storagecoats branch on my repo to your own repo.

Nnelg:
It's a longer project, and I can't really start it until finals are over; so don't feel bad about coding this one now if you want to.

If nothing else, this method is known to work. My planned code is completely experimental.

Iamgoofball:
code/modules/clothing/suits/jobs.dm:11:error: blood_overlay_type: undefined var

code/modules/clothing/suits/jobs.dm:61:error: blood_overlay_type: undefined var

code/modules/clothing/suits/jobs.dm:70:error: blood_overlay_type: undefined var

code/modules/clothing/suits/labcoat.dm:6:error: blood_overlay_type: undefined var

PR #471, opened 2014-04-20
Adds skull tank top, fixes error with russian longsleeve

ghost:
That error where the longsleeve russian shirt was moved down a pixel is fixed.

Skull tank top is added and it's for all the greyshirts out there that need to tell everyone that the spiderlings are craaaawlliiiing innnnn their skiinnnnn

clip 2014-04-19 at 11 11 46

these wouuunds they wiiiil not heauuulllll


Atskadan:
(thumbs up)

PR #473, opened 2014-04-20
AIs. New core, AI death specific.

Niknakflak:
https://github.com/tgstation/-tg-station/pull/3462

NTstation verson.
-Adds Heartline AI core.  NTstation has all other cores.
-Adds death core icon for Heartline and Robert house
-Adds code to specify specific death icons for cores if one is available. Heartline goes Flatline when dead.  Robert house goes connection lost when dead.  If no death core is specified, it will default to blue screen of death.  To add a death core, simply make a sprite of what the core would look like when dead and name it ai-(name)_dead.  
-Fixes the Helios core having a rouge pixel frame on the right.


RemieRichards:
It's tested Danno.

Evidence: Me and Wb and Niknak in Nanobus

PR #483, opened 2014-04-25
You can log into ID computers with only access_change_ids again

ghost:
It wasn't very intuitive for players who thought access_change_ids was enough to use the computer. HoPs should rest easy now.


Atskadan:
yay

PR #499, opened 2014-04-27
New Human Mobs will have the appearance of the client's preferences

ghost:
New human mobs created via Admin Simple Transform, Admin Spawn Antagonist, and Wizard Apprentice functions will take on the appearance of the players' preferences, instead of the boring default Bald, White Male.

Also fixes an minor issue with using transfer_to() a new observer mob.


Atskadan:
end bald white male patriarchy 2558

PR #504, opened 2014-04-28
Adds "OnEquip()" an Item proc that runs code when the item is equipped

RemieRichards:
Does NOT work on pickup, because that'd be stupid.

Current uses implemented with this PR:
- Clown mask honks
- Jumpsuits play 'zip.ogg' for zipping up your jumpsuit
- New "Shamedora" fedora that adds bruteloss upon wearing


RemieRichards:
Nevermind, a proc DOES exist, I just missed it.

PR #508, opened 2014-04-30
More Common NPC Blob Event

ghost:
-Adds a blob event variation that is more common and no round time delay. Spawns a weaker, NPC blob core.
-Brand Intelligence event added to admin secrets panel


Atskadan:
sure

PR #527, opened 2014-05-07
Makes traitor~chan uguu 15 players min instead of 20

ghost:
Fuck boring lowpop changeling rounds.


GunHog:
Paprika for Funcoder 2014 :+1:

PR #532, opened 2014-05-07
Introducing the Hostile Lockdown ability for AI Malfunction!

GunHog:
Grants a Malf AI a new ability - Hostile Lockdown, costing 20 CPU time.
- This ability activates all firelocks
- Closes all blast doors and shutters
- Closes and shock-bolts all airlocks that the AI can control, crushing anyone standing in a doorway when it happens.
- Sets the screen display to "Lockdown"

The lockdown can be reversed using the "Disable Lockdown" verb.
- Lifting the lockdown opens everything that was locked down by the AI. (Airlocks close again shortly after)
- Lifting the lockdown restores airlocks to a safe and functional setting.

Justification for this ability provided here: http://www.ss13.eu/tgdb/tg/latest_stats.html#mode_AI_malfunction

Discussion and poll relating to this update located here: 
http://tgstation13.org/phpBB/viewtopic.php?f=14&t=284

/tg/ PR version found here: https://github.com/tgstation/-tg-station/pull/3421

### DIFF
--- a/code/game/gamemodes/malfunction/Malf_Modules.dm
+++ b/code/game/gamemodes/malfunction/Malf_Modules.dm
@@ -72,9 +72,7 @@
 
 	for(var/obj/machinery/firealarm/FA in machines) //activate firealarms
 		spawn()
-			if(FA.lockdownbyai == 0)
-				FA.lockdownbyai = 1
-				FA.alarm()
+			FA.alarm()
 	for(var/obj/machinery/door/poddoor/BD in world) //Close blast doors!
 		spawn()
 			BD.close()
@@ -105,9 +103,7 @@
 
 	for(var/obj/machinery/firealarm/FA in machines) //deactivate firealarms
 		spawn()
-			if(FA.lockdownbyai == 1)
-				FA.lockdownbyai = 0
-				FA.reset()
+			FA.reset()
 	for(var/obj/machinery/door/poddoor/BD in world) //Open blast doors!
 		spawn()
 			BD.open()
--- a/code/game/gamemodes/malfunction/Malf_Modules.dm
+++ b/code/game/gamemodes/malfunction/Malf_Modules.dm
@@ -80,7 +80,7 @@
 			BD.close()
 	for(var/obj/machinery/door/airlock/AL in world) //shock-bolt airlocks
 		spawn()
-			if(AL.canAIControl())
+			if(AL.canAIControl() && !AL.stat) //Must be powered and have working AI wire.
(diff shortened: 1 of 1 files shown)


Neerti:
If I were the one making this PR, I'd be bold and have it be a free ability, since doors aren't very great at stopping people and there's other exits, but you're not me, so :+1:.

Zelacks:
shock bolts all doors, that is so strong though

GunHog:
That is why I set the cost so low. Rememer that you are indeed shock-bolting all airlocks at once! You will not stop engineers, assistants/cargo, certain heads or the clown, but everyone without tools or gloves is going to have a hard time with things.

Also, the lockdown can be reversed and set again, so if the crew is not cutting airlock power or AI control wires, airlocks that have been hacked open can and will be shock-bolted again!

Also..what is wrong with Travis? This code is an exact duplicate of the /tg/ version. I compiled and tested both with no issues...

GunHog:
That is just it. The code is not broken. I do not know what to do.

RemieRichards:
Can you please swap the usr's for src's Because usr is terrible and should never be used anywhere even for messages (And the fact you're checking usr's stat var)

GunHog:
Swapped them. Tested and confirmed working as intended. Also, checking for a dead AI is important because all of the verbs without one can be used while the AI is dead. We cannot have a dead AI shock-bolting all the airlocks.

RemieRichards:
Oh no, I wasn't contesting the stat check, of course it's needed, I was saying it's another reason not to use usr. but it worked out anyways.

GunHog:
Thanks to all who helped test! It was quite fun! Found a bug where unpowered airlocks could be unbolted and otherwise modified by the lockdown powers. Cutting AI the control wire, the power wires, or the controlling APC unit will prevent them being affected by the lockdown or its counterpart.

Atskadan:
this is why danno

Atskadan:
but more yeses

GunHog:
Any specific _reason_ they find it op? The ability to reset it? That it shocks doors as well? Are we dealing with a large number of complaints, or a vocal minority? I am open to change if there is truly a problem. I would like to gather more results. Two rounds for something quite new is not that helpful.

If it really, truly is OP, I will adjust or remove it myself. I would like to be sure it is not an overreaction to a new feature, is all.

Zelacks:
I mean people were blasting this for it being too tame and ineffective. Because the ai has the abilitiy to do this anyway, but at slower speeds.

PR #537, opened 2014-05-08
Changes the Boozepwr of drinks

Atskadan:
so you don't become a stuttering mess so quickly

these values are easily changed so feel free to suggest different ones

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents.dm
+++ b/code/modules/reagents/Chemistry-Reagents.dm
@@ -2849,7 +2849,7 @@ datum
 				return
 
 /*boozepwr chart
-55 = non-toxic alchohol
+55+ = normal drinks
 45 = medium-toxic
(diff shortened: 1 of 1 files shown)


ghost:
can you add breathalyzers for sec so we can brig people for being too drunk on the job

vista-:
To be honest, I haven't encountered any problems with booze strength, you have to drink like, more than 75 units of beer to become a stuttering mess.

Atskadan:
yeah a single drink of pretty much any alcohol out of the bottle will make you drunk
honestly i dont think these values are even enough

Petethegoat:
Vistapowa it's delayed basically. Pretty sure one sip of spacebeer will fucking wreck you forever, when it kicks in after a minute or two.

Like how if you eat a single donut at roundstart you're fucked fat forever gg no re good luck walking that off fatty

PR #538, opened 2014-05-09
SSU Updates

ghost:
-Added an auxiliary storage compartment in SSUs to store one item
-Moved CE's magboots and captain's jetpack into them
-Updated it so that any mob can be placed inside the SSU now. Including Corgis.
-Fixes engineering SSUs


ghost:
> burning corgis alive

clip 2014-05-08 at 10 44 57

Also I'm going to be overhauling the nuke shuttle to use this and putting the nuke op jet harnesses in these eventually, good work.

PR #540, opened 2014-05-09
Adds West and South directions to moving space

Tomeno:
Icons edited by WJohnston.


WJohn:
Wonderful! This will help all mappers in the future! No longer do we have to have shuttles and pods facing north or east!

PR #541, opened 2014-05-10
Attention! Code Delta!

GunHog:
- When a nuke is armed, it will set alert level Delta.
- This includes the nuke used by Operatives, as it is just a nuke stolen from another station (and it is why the Ops need to steal the disk to use it).
- If the nuke is disarmed, it will drop the alert level to whatever it previously was before the timer was triggered.

Exact copy of the Nuke code used by /tg/, now brought to you!
The results from /tg/ appear to cause increased panic during Nuke Ops rounds! In addition, Admins have used it to force Code Delta alerts!


ghost:
I like it, 60 seconds of 'OH FUCK GRAB A SUIT AND RUN TO AN AIRLOCK' would make nuke op way more fun.

PR #543, opened 2014-05-10
Adds atmos coats to atmos wardrobe lockers and moved pinpointer to HoS office

ghost:
Not exactly sure why the HoS never had a pinpointer when it's practically his job to babysit the cap.

Also atmosia coats yeah!!!


GunHog:
Remove the Captain's if you are giving it to the HoS. There should only be one of these so that the strategy of spacing the pinpointer is not made two times harder for Stealth Ops and Malfunctioning cyborgs.

ghost:
No PR is safe from gun hog's keen AI regular eye

ghost:
The cap should be aware of the recent change since the note is literally right next to the disk so here's hoping people don't faff about it suddenly being missing!

Atskadan:
yes please i was so cold without my atmos coat ;n;

PR #547, opened 2014-05-11
Blocks non-carbon mobs from most click-drag functions

ghost:
This means that they can no longer:
-Strip people
-Give people things they shouldn't have (eg cyborg modules)
-Search through people's backpacks and boxes
-Buckle people in


Atskadan:
ok

PR #550, opened 2014-05-11
Adds Dominator Guns [Poll Ongoing]

RemieRichards:
## Dominators:
- A Unique gun that requires the user to have Security access to fire it.
- The Gun will NOT fire if the user is a criminal
- The Gun will NOT stun the target if the target is not a criminal (If you're shot with this, and it stuns you, you literally can't scream "shitcurity" because you need to be in the Sec Records listed as a criminal)
- The Beepsky sec record trick DOES work with Dominators
## Specs:
- Double the Taser's Stun
- Double the Taser's Weaken
- Double the Taser's Ammo
## Code Extras:
- Adds a human helper proc "IsCriminal()" which checks for valid criminal ranks in Sec Records

<b>Thread:</b> http://tgstation13.org/phpBB/viewtopic.php?f=15&t=387


Atskadan:
needs unique sprite and kill modes pls

RemieRichards:
Look at the tile miggles

> [Needs Sprites]

Also the Kill modes will probably not get in or will have to come later, due to them requiring extra balance etc.

Atskadan:
it should honestly be an admin item first, then balanced
i dont think know about giving security guns that are just references to anime

RemieRichards:
@Atskadan The Poll has shown the players are (majority) in favor of the Unique concepts of the Dominator in it's Current implementation, I think that's fine enough to add it/run a test of them in the code for a while.

Atskadan:
if youre not going to make it a direct reference to psycho-pass, you should change the name
it would be dumb to call it a dominator and it doesnt even look like one

RemieRichards:
Jeez miggles a name is a name. but _ok_ I guess I'll look for a new name for them.

Neerti:
> Gar Glasses
> 
> No Refernces

Atskadan:
its not >no references
its >no inaccurate references
*this isnt a rule btw

RemieRichards:
Well the only included references are:
- The Name
- It's "officer" only characteristic
- It's only criminals are shot characteristic

those are the 3 references, all are accurate.
The point is you want something more like the dominator, and considering that will likely never get added, you'd rather this "half way" stage is renamed, Correct?

Atskadan:
Well I mean the whole concept is based off of the ones from Pyscho-Pass, it's just not complete. I just don't personally like inaccurately referencing something. You don't have to rename it if you don't want.
Like, the e-sword could easily be renamed to "lightsaber." But that would be dumb, because it's just based on the same concept. It's an obvious reference to Star Wars, but it's still called something else.

Iamgoofball:
how about we allow creative license @Atskadan

Atskadan:
> *this isnt a rule btw

did you miss the part where everything i say is opinions

RemieRichards:
Experiment failed, not wanted/not balanced fairly.

Atskadan:
@RemieRichards just make it admin spawnable pls

RemieRichards:
Maybe later. My branches are all tied up.

PR #552, opened 2014-05-12
Engineers know Door and APC wires from Round start [Merge]

RemieRichards:
At Roundstart an Engineer (CE, Engi or Atmos)'s Notes are filled with the Wires of APCs and Doors.
Adds an Implant to Syndicate Uplinks to grant them knowledge of Door wires (1 TC cheaper than an Emag because it requires the user to obtain tools)

Codewise:
Adds 3 procs.
- Get's the name of the object by it's wire datum
- Deciphers the Randomised Wire list
- Translate the Deciphered Bitflags into readable english

The SolveWires() proc outputs a string (Text) to an associative list based on the Wire datum passed to the SolveWires() Proc, This text string is valid where all others are, such as Memory procs.

Eg; all_solved_wires[/obj/machinery/door/airlock] contains the translated string from SolveWires()

### DIFF
--- a/code/game/objects/items/weapons/implants/implantknowledge.dm
+++ b/code/game/objects/items/weapons/implants/implantknowledge.dm
@@ -1,0 +1,21 @@
+
+/obj/item/weapon/implant/wire_knowledge/implanted(mob/M)
+	if(!M.mind)
+		return 0
+	M.mind.store_memory("If you're reading this, Blame the nearest online Admin")
+	return 1
+
+/obj/item/weapon/implant/wire_knowledge/door/implanted(mob/M)
+	if(!M.mind)
+		return 0
+	M.mind.store_memory(all_solved_wires[/obj/machinery/door/airlock])
+	return 1
+
+/obj/item/weapon/implant/wire_knowledge
+	name = "Wire Knowledge: Blank"
+	desc = "This doesn't do ANYTHING"
+
+
+/obj/item/weapon/implant/wire_knowledge/door
+	name = "Wire Knowledge: Doors"
+	desc = "Grants you knowledge of the NT Door and Airlock wiring systems"
--- a/NTstation13.dme
+++ b/NTstation13.dme
@@ -586,6 +586,7 @@
 #include "code\game\objects\items\weapons\implants\implantchair.dm"
 #include "code\game\objects\items\weapons\implants\implanter.dm"
 #include "code\game\objects\items\weapons\implants\implantfreedom.dm"
+#include "code\game\objects\items\weapons\implants\implantknowledge.dm"
 #include "code\game\objects\items\weapons\implants\implantpad.dm"
 #include "code\game\objects\items\weapons\implants\implantuplink.dm"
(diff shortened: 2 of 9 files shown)


ghost:
2 is pretty fair, basically makes gloves useless while making a good alternative to the emag. Excellent rob. Might want a [WIP] tag on this.

NTstation13.dme:589:error: unable to open "code/game/objects/items/weapons/implants/implantknowledge.dm".

RemieRichards:
It isn't WIP, but I think I forgot to commit that file, DOH.

Giacom:
I'm crying.

RemieRichards:
@Giacom I'm sorry, I tried my best to make it "one size fits all" (as with the SolveWires() proc) but I still needed the Translate proc to be genetic and overridden in the child datums.

Your code babby was beautiful.

code/game/objects/items/weapons/implants/implantknowledge.dm
```diff
@@ -0,0 +1,21 @@
+
+/obj/item/weapon/implant/wire_knowledge/implanted(mob/M)
+	if(!M.mind)
+		return 0
+	M.mind.store_memory("If you're reading this, Blame the nearest online Admin")
```
Zelacks, line 5: Good joke, but programming is no joke! Basically the feedback is nonsensical garbage, which is cute but isnt practical. You will either puzzle the player or the admin when the player decides to consult the admin.
Give it meaningful feedback or remove it. (Or dont I am not an enemy please dont hate me i am just trying to help)

Of course I am aware that I really have no idea what I am talking about probably and its something like the mob needs to have memory stored or it breaks

RemieRichards, line 5: No no, I could simply have it return, the message isn't required, but given this item being the base knowledge implant with no design function besides adding a message to the player's mind, it should never be accessed, hence the "blame an Admin" because only an admin could get a blank implanter, faff with an implant case to add the implant to it, and then load the Implanter.

But if it's warranted I can remove it with no issues.

Zelacks, line 5: Yeah in that  case, its probably cool to keep the message. But it should read something along the lines of (parapharsing this is probably terrible change it to something better than what I say) "An admin has spawned the incorrect class of implanter! Please tell them to spawn a "/obj/item/weapon/implant/wire_knowledge/door" instead."
It could also be an admin message or something, but then you might get something fuckity going on

ghost:
Hey maybe engineers will actually cut malf AIs access out of APCs and doors instead of letting people get crushed in them over and over

code/game/objects/items/weapons/implants/implantknowledge.dm
Atskadan, line 5: i like zelacks idea

RemieRichards, line 5: Really? Considering this dummy implant can't even be used without MASSIVE fuckery like I mentioned, an Admin will immediately realise they need to spawn an implanter with a proper one already in or the syndi kit box.

PR #562, opened 2014-05-14
ponytail 3 minor shading update

ghost:
I couldn't resist updating out the blob shading on this one and a couple other pixel adjustments to make it less symmetrical and boring looking.


Atskadan:
before/after?

ghost:
ss 2014-05-14 at 04 33 59

ss 2014-05-14 at 04 34 42

The worst part is if you look at the 'facing north' sprite for ponytail 3 it's actually shaded but whoever did it got really fucking lazy with the west/east/south sprites so I finished what they started so it doesn't look like a blob of pixels anymore. It also looks 'fuller' now, like a messy nape of neck mini ponytail for hipsters or antonio banderas

PR #563, opened 2014-05-14
Sec borg hailer + some other stuff

ghost:
HALT HALT HALT HALT 

has been replaced with

HALT HALT HALT

...because the quad halt was a little too long in my opinion.

Also, the borg sec hailer mask starts at first position on the restrictor so it will not yell any mean things at the humans!

If someone eventually makes that 'integrated sec hud' thing for the sec borg module you should also integrate the hailer.


Atskadan:
why does aggressiveness = 1 mean that it isnt aggressive
that is backwards sense

Neerti:
> Implying the code is meant to make sense.

Atskadan:
>not knowing how to greentext

PR #567, opened 2014-05-15
rev up those droprates

ghost:
I've heard mixed opinions on my buff to droprates, piggybacking on ergo's proposed buff (and subsequent nerf to rnd prices) but the opinion has mostly been 'not enough'. This is a tiny bit of an update and I'd like to hear your opinions since it's mostly value editing and really easily changed.

tl;dr: mining should be mining again and not monster hunter laggy 2d edition

(also diamond no longer spawns in the primary chunk of the east asteroid at all)

### DIFF
--- a/code/modules/mining/mine_turfs.dm
+++ b/code/modules/mining/mine_turfs.dm
@@ -73,7 +73,7 @@
 	name = "Mineral deposit"
 	icon_state = "rock"
 	var/mineralSpawnChanceList = list("Uranium" = 20, "Diamond" = 5, "Gold" = 12, "Silver" = 18, "Plasma" = 25, "Iron" = 40, "Gibtonite" = 2/*, "Adamantine" =5*/, "Cave" = 2)//Currently, Adamantine won't spawn as it has no uses. -Durandan
-	var/mineralChance = 15  //means 15% chance of this plot changing to a mineral deposit
+	var/mineralChance = 20  //means 20% chance of this plot changing to a mineral deposit
 
 /turf/simulated/mineral/random/New()
 	..()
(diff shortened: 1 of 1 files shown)


Iamgoofball:
someday we need to find urist and drag his lazy ass back over here and make him finish adamantite

code/modules/mining/mine_turfs.dm
```diff
@@ -129,7 +129,7 @@
 /turf/simulated/mineral/random/low_chance
 	icon_state = "rock_lowchance"
 	mineralChance = 5
-	mineralSpawnChanceList = list("Uranium" = 5, "Diamond" = 2, "Gold" = 5, "Silver" = 10, "Plasma" = 15, "Iron" = 35, "Gibtonite" = 1)
+	mineralSpawnChanceList = list("Uranium" = 5/*, "Diamond" = 2*/, "Gold" = 5, "Silver" = 10, "Plasma" = 15, "Iron" = 40, "Gibtonite" = 1)
```
Iamgoofball: i dunno about this man

ghost: It seems silly to have diamond on the east rock at all but I can revert it if you really want it to be #rare as fuck

Atskadan:
it still seems low
only diamond, plasma, and iron are changed
and only by increments of 5

ghost:
now there's a 20% chance of a mineral turf containing a mineral, compared to the old 10%

this is the 'general' drop rate so there should be a notable increase in minerals around now

GunHog:
Yaaaaaaay! :+1:

ghost:
It has a lot to do with how the map is layed out now because the old way minerals works was a little wonky but this should be close if not better than old mining but I really need to know if miners are coming back with three full stacks of diamond after 10 minutes once this has been tested. Looking at it in game it looks pretty tame, but you never know.

ghost:
Merge this for confirmed fun

PR #575, opened 2014-05-17
AI Sensor Augmentations! [Ready]

GunHog:
Discussion and poll here: http://tgstation13.org/phpBB/viewtopic.php?f=14&t=440
- Security Augmentation
  This function gives the AI the ability to identify the job and security status of any human crew member it sees. Unlike the cyborg or human versions, it strictly read-only. The AI is not able to set a human's security status this way, nor can it detect any implants. Its function is equal to that of the one used by pAI units.
- Life Signs Augmentation
  Perhaps the most useful of settings for Asimov AIs, this augmentation gives the AI a read-only medical HUD. Identical to the pAI function, it will provide data on wellness and infection status.
- Light Amplification Augmentation 
  An extremely helpful function to have during a power or lighting outage, it will function as NVG do for humans. The AI will be given vision in darkness such that it may remain operational in areas it is most critically needed.

*Note, This does not increase the AI's vision range! The AI has always been able to see and use machines in darkness. Blind spots as a result of damaged or missing cameras will remain.

### DIFF
--- a/code/modules/mob/living/silicon/ai/ai.dm
+++ b/code/modules/mob/living/silicon/ai/ai.dm
@@ -51,6 +51,10 @@ var/list/ai_list = list()
 	var/last_paper_seen = null
 	var/can_shunt = 1
 	var/last_announcement = "" // For AI VOX, if enabled
+	var/sensor_mode = 0 //Determines the AI's current HUD.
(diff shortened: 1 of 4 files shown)


GunHog:
Keeping Nightvision mode unless sufficiently requested otherwise!

PR #580, opened 2014-05-17
Fixes Immovable Rod

ghost:
-Rod no longer gets moved into nullspace after hitting a few objects
-Rod no longer gets stuck in the rare case it hits a shuttle at just the right angle. It now explodes.
-Rod now makes a lot of audio clamor when breaking through things in addition to the text


ghost:
I hope it's the disposals bang noises.

PR #595, opened 2014-05-19
Several guns changed to right handed + 9mm buff

ghost:
Included weapons are:
Normal shotgun
Combat shotgun
L6 SAW
Glock (revert of a previous flip I did, my bad)

Since there's no -real- consistency with the way guns face this should work. Shell ejection ports on shotguns are now on the right side of them. L6 saw 'belt' feeds into the weapon from the left of the gun and ejects spent casings out the right, which is how most 'merican light machineguns feed.


ghost:
9mm now does 20 brute. For comparison, the slightly larger but less capacity M1911 does 20 brute as well but also has a stunning affect.

Stech still does 30 for traitoring.

ghost:
merge this pl0x

PR #597, opened 2014-05-20
You can now punch silicons.

ghost:
It doesn't hurt them unless you're hulked.


Atskadan:
end silicon abuse 2556

PR #598, opened 2014-05-20
Grants Silicons the ability to hack bots! [Port from /tg/][Poll in progress]

GunHog:
A silicon (AI, cyborgs) is now able to hack any robot through that robot's interface.

A hacked robot behaves mostly as an emagged one does.

Hacked robots cannot be deactivated via interface. _If it was manually emagged, it cannot be stopped except by destruction._ If a silicon hacked the bot, it must be reset by a silicon before it can be shut down.

Manually emagged robots cannot be restored by silicons. They must be destroyed instead.

The following robots are affected:
- Medibots (including Mysterious versions)
- Securitron
- ED-209 (including laser tag versions!)
- Floorbot
- Cleanbot

This is adapted from the /tg/ version, but more-or-less the same code! Poll link below!
http://tgstation13.org/phpBB/viewtopic.php?f=14&t=472

### DIFF
--- a/code/game/machinery/bots/bots.dm
+++ b/code/game/machinery/bots/bots.dm
@@ -13,6 +13,7 @@
 	var/brute_dam_coeff = 1.0
 	var/open = 0//Maint panel
 	var/locked = 1
+	var/hacked = 0 //Used to differentiate between being hacked by silicons and emagged by humans.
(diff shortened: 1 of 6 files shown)


GunHog:
https://github.com/NTStation/NTstation13/pull/601 includes my original code of this, so if that one goes through, this can be closed.

PR #599, opened 2014-05-20
[don't merge just yet]You can now climb on tables and racks.

Iamgoofball:
seriously why was this not done before

hulks can't climb on tables or racks, they just smash them instead because hulk smash or whatever

### DIFF
--- a/code/game/objects/structures/tables_racks.dm
+++ b/code/game/objects/structures/tables_racks.dm
@@ -476,7 +476,12 @@
 			new /obj/item/weapon/table_parts(loc)
 		density = 0
 		qdel(src)
-
+	else
+		if(!user.loc == src.loc)
+			user << "You climb onto the table."
+			user.loc = src.loc
+		else
+			return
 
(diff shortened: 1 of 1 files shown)


Atskadan:
this is really bad
tables and racks are used as pseudowalls all over the place on the map
climbing on them will lead to all sorts of problems
>climb into HoP office
>climb into kitchen/bar
>climb into medbay/chemistry
>climb into arrivals checkpoint
etc etc

Iamgoofball:
if an area is secure enough this is a problem

it should of had windoors on the tile w/ the table to start with

Iamgoofball:
If tables were climable at the start this wouldn't be an issue.

If you give me a list of areas I can beef up table security and add windoors.

Atskadan:
what benefits are there in climbing over tables?
do you really think adding windoors to everything is a solution to a problem that shouldnt even exist?

RemieRichards:
Don't like it :-1: My reasons are the same as the others.

PR #613, opened 2014-05-23
Pants and leather jackets!

JStheguy:
-Adds Khaki pants, Jeans, and Track pants.
-Adds a leather jacket that is made using the biogenerator.
-Adds a casual clothing closet (Has bandanas, beanies, pants, and jackets) to arrivals shuttle and the gambling room in the bar.


JStheguy:
ss13shuttle
Here, picture of leather jacket and the jeans.

PR #614, opened 2014-05-23
fixes invisible stech mag

ghost:
i forgot stechkins use the 9mm mag icon when i altered the icon state names for the glock but it's fixed now


Atskadan:
wow youre so bad at things stop being bad, baddie

PR #615, opened 2014-05-23
Colored and clear glasses

ghost:
ss 2014-05-23 at 04 55 58

Could someone do me a favor and code these to work with the washing machine? Not sure how that works but here you go, the items, and all the crayon colors, with a few extra.


RemieRichards:
The washing machine is ancient TERRIBLE code, not even erro who wrote it is willing to touch it.

ghost:
Bursts into treats.

RemieRichards:
> treats

Atskadan:
that is the joke

RemieRichards:
_flails arms_

ghost:
it has something to do with item_color as in the green glasses would be item_color and washing glasses just makes them use the icon state or 'item color' greenglasses, it's really weirdly complicated though, not sure if anyone can decode it but i don't have the time currently

PR #616, opened 2014-05-24
removes inhale from soup eating

ghost:
Why would you inhale soup/stew


Atskadan:
eating quickly

ghost:
then you'd choke

PR #618, opened 2014-05-24
The Belt of Holding!

GunHog:
Adds the Belt of Holding!
- It is a effectively a backpack for your belt!
- It has the same ability to create a singularity that the bag of
  holding does.
- Like the Bag of Holding, it will disrupt teleports.
- Same tech cost of a Bag of Holding, but at a lower cost, as it is an
  inferior item.

Sprite by MrSnapwalk.


Atskadan:
neato, and nice sprite

PR #620, opened 2014-05-24
Changes parapen to 4tc

ghost:
What it says on the tin.

Poll here:
http://tgstation13.org/phpBB/viewtopic.php?f=5&t=526


ghost:
Based on the resounding feedback and poll results on the forum I'm going to call this one safe to merge. If it causes significant issues down the line I'll look at it again but I'm pretty sure that won't be a thing.

PR #635, opened 2014-05-26
Fixes borg arm icon

ghost:
Borg arm icon was being interfered with by this unused 'overdrive' item giving it the icon_state = "shock" but I removed that and now it works properly.


ghost:
ss 2014-05-25 at 11 32 44

PR #647, opened 2014-05-28
Surplus clothing crate 

JStheguy:
A supply crate with 10 articles of randomly chosen articles of clothing, costs 15 supply points.


Psyentific:
Seems a bit expensive.

Atskadan:
its 15 points

PR #666, opened 2014-06-01
Posters Thing

JStheguy:
Adds new posters and a subtype system that allows for posters that
aren't part of the contraband posters specificly.
image


ghost:
I love this.

PR #685, opened 2014-06-04
Adds zoom scope to ion rifle [wip]

ghost:
This is really bad and mostly a joke but it's half working at the moment. Things that need to be implemented/fixed:
- Unzooms when it leaves your active hand/is put on backpack/dropped
- Unzooms when you move, doesn't work at the moment
- HUD disappears after zoom out, might remove the HUD minimalizing on ADS entirely if this can't be fixed
- Zooms when you press z/click on it in active hand (use it in other words) for easy toggling


Neerti:
Will this let me 360 noscope all the rogue borgs?

RemieRichards:
I don't like it, but the code is fine.

code/modules/projectiles/guns/energy/special.dm
```diff
@@ -9,6 +9,46 @@
...
+This is called from
+modules/mob/mob_movement.dm if you move you will be zoomed out
+modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
+*/
+
+/obj/item/weapon/gun/energy/ionrifle/verb/zoom()
+	set category = "Object"
+	set name = "Use Scope"
+	set popup_menu = 0
+	if(usr.stat || !(istype(usr,/mob/living/carbon/human)))
```
Iamgoofball, line 27: > usr

code/modules/projectiles/guns/energy/special.dm
```diff
@@ -9,6 +9,46 @@
...
+		usr << "You are unable to focus down the scope of the rifle."
```
Iamgoofball, line 28: > usr

ghost:
@goof that was baycode, how do we do it differently here? span classes for the messages?

RemieRichards:
@paprka usr has nothing to do with messages. usr should NEVER be used in a proc or a verb.
you really need to type cast usr into a /mob/living/carbon/human and then use that.

ghost:
too lazy to leave this shameful copy paste up

PR #689, opened 2014-06-05
You can now grow gatfruit.

Iamgoofball:
Honk.


Neerti:
You can reload russian revolvers w/ normal revolver ammo and it will almost work the same so :-1:

Iamgoofball:
well fuck

Yakk:
There is only one russian revolver and it does less damage when fired at other people than normal revolvers.

PR #698, opened 2014-06-07
admin permissions

Iamgoofball:
why people w/o +ADMIN could use admin stuff, I don't know.

this makes shit that should of had +ADMIN actually require +ADMIN. by default someone in the TXT only gets deadmin self, investigate, and the jumps now. if you don't have +ADMIN you don't need anything other then those.

### DIFF
--- a/code/modules/admin/admin_memo.dm
+++ b/code/modules/admin/admin_memo.dm
@@ -6,7 +6,7 @@
 	set name = "Memo"
 	set category = "Server"
 	if(!ENABLE_MEMOS)		return
-	if(!check_rights(0))	return
+	if(!check_rights(R_ADMIN))	return
 	switch(task)
 		if("write")		admin_memo_write()
 		if("show")		admin_memo_show()
@@ -14,6 +14,7 @@
 
 //write a message
 /client/proc/admin_memo_write()
+	if(!check_rights(R_ADMIN))	return
 	var/savefile/F = new(MEMOFILE)
 	if(F)
 		var/memo = input(src,"Type your memo\n(Leaving it blank will delete your current memo):","Write Memo",null) as null|message
@@ -31,6 +32,7 @@
 
 //show all memos
 /client/proc/admin_memo_show()
+	if(!check_rights(R_ADMIN))	return
 	if(ENABLE_MEMOS)
 		var/savefile/F = new(MEMOFILE)
 		if(F)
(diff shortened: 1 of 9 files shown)


code/modules/admin/admin_memo.dm
```diff
@@ -6,7 +6,7 @@
 	set name = "Memo"
 	set category = "Server"
 	if(!ENABLE_MEMOS)		return
-	if(!check_rights(0))	return
+	if(!check_rights(R_ADMIN))	return
```
MrsTonedOne, line 9: Wouldn't this make it so people who are admin observers or otherwise in admin.txt with no flags couldn't check what the memo is?

code/modules/admin/admin_memo.dm
```diff
@@ -31,6 +32,7 @@
 
 //show all memos
 /client/proc/admin_memo_show()
+	if(!check_rights(R_ADMIN))	return
```
MrsTonedOne, line 35: Here you are doing the same thing, preventing people in admins.txt from seeing the admin memo if they don't have +admin, ie, admin observers.

code/modules/admin/admin_verbs.dm
```diff
@@ -1,21 +1,14 @@
 //admin verb groups - They can overlap if you so wish. Only one of each verb will exist in the verbs list regardless
 var/list/admin_verbs_default = list(
-	/client/proc/toggleadminhelpsound,	/*toggles whether we hear a sound when adminhelps/PMs are used*/
```
MrsTonedOne, line 3: So anyone with a holder or otherwise is in admins.txt without +admin or other flags can hear ahelps, they can see them, but you now prevent them from being able to silence the annoying sound they make.

And you prevent them from being able to reply to the- oh wait, you don't do that, because none of the changes you made are to topic.dm, thus all of the admin links that do these same things still work for admins without +admin.

Iamgoofball, line 3: I'll move that one back to default.

code/modules/admin/admin_memo.dm
Iamgoofball, line 9: I'll fix it

Atskadan:
What you're doing is changing the structure of admin rights because people complained you were misusing your +ADMIN flag that you needed so you could <i>spam sounds</i>.
I'm sorry, no. This isn't warranted and doesn't really benefit anyone. 
You know what you do if people tell you to stop using buttons? <i>Stop using them.</i> That INCLUDES playing sounds - regardless of if there is a mute button or not. You're not in any position to be arguing against people telling you you're wrong.

Iamgoofball:
I never had +admin

Atskadan:
Yes, I was mistaken when I said that. That's besides the point.

PR #711, opened 2014-06-12
port for floaty spacemen

ghost:
ports https://github.com/tgstation/-tg-station/pull/3885

the best feature of all time

also fixes magboot / adv magboot mob sprite only having one boot light up green (never been fixed apparently?)

based giacom deliver us unto immersion


RemieRichards:
Is it really the best idea to port this? It'll be added with the big merges from /tg/ (presuming we keep doing those)

ghost:
RR i didn't get the impression it would be a regular thing, the big porting, plus the last time it happened it had a host of bugs, so next time i'd rather if it was tested and pruned more thoroughly. But like danno said, we want floaty spacemen now, it actually makes a host of difference for immersion and i love it a lot.

RemieRichards:
Alright, I'll budge for now, but Whoever merges from /tg/ do NOT complain about having to work around this.

> kun

:/

PR #727, opened 2014-06-19
Blob is 20 required players

ghost:
We never get to play blob on this server nobody fucking readies ;_;

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -8,7 +8,7 @@
 /datum/configuration
 	var/server_name = null				// server name (for world name / status)
 	var/server_suffix = 0				// generate numeric suffix based on server port
-	var/lobby_countdown = 120			// In between round countdown.
+	var/lobby_countdown = 180			// In between round countdown.
 
 	var/log_ooc = 0						// log OOC channel
 	var/log_access = 0					// log login/logout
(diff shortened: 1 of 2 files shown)


ghost:
Lobby timer is now 3 minutes because wow nobody readies (including me)

Like shit I can make a poll on this if you'd like but holy shit every time I come back if I go get water or something the round has already started with a prob of traitor only and it sucks

Atskadan:
no this is really good

GunHog:
MERGE NOW PLZ :+1:

Iamgoofball:
uhm

the lobby timer should of been in another pr

but okay

GunHog:
Sounds like a config thing. SoS could probably change it on the fly. I am fine with it, but I do not think one minute or ten would make people ready up.

Iamgoofball:
I don't, but it probably should of been in a separate PR if someone wanted to find when this was added and what PR it was in. They'll just see "oh this only touches blob, it isn't this one" because of the title

GunHog:
Seems an easy issue to fix. Rename the PR!

Atskadan:
youre one to talk about hiding things in pr's

Iamgoofball:
Honk

code/controllers/configuration.dm
```diff
@@ -8,7 +8,7 @@
 /datum/configuration
 	var/server_name = null				// server name (for world name / status)
 	var/server_suffix = 0				// generate numeric suffix based on server port
-	var/lobby_countdown = 120			// In between round countdown.
+	var/lobby_countdown = 180			// In between round countdown.
```
Giacom, line 11: You need to change the config setting itself too, in the .txt. You will then have to tell SoS about the change so he can change it on Artyom.

PR #729, opened 2014-06-19
[WIP] Porting bay holsters and webbing

ghost:
Tiecode a shit. One day I'll make this work, and where miggles failed to port labcoat storage I will succeed.


code/modules/clothing/clothing.dm
```diff
@@ -258,42 +270,91 @@ BLIND     // can't see anything
 		if(3)
 			usr << "Its vital tracker and tracking beacon appear to be enabled."
 	if(hastie)
-		usr << "\A [hastie] is attached to it."
+		usr << "\A [hastie] is clipped to it."
```
ghost, line 273: don't mind this i will turn it back

ghost:
anyway when a code savvy person is here this is basically a direct port so figure out what's wrong with internal storages and stuff okay :+1: 

afaik some of the main problems are:
1. you can't pick up holsters at all
2. webbings work for storage and everything else but you can't take things out of storage, once this is solved making labcoats/jackets have pockets should be pretty simple
3. this basically breaks tie icons and stuff using this new overlay code which i don't know how to add here or revert to working properly

code/modules/clothing/clothing.dm
```diff
@@ -258,42 +270,91 @@ BLIND     // can't see anything
...
+					V.show_message("\red [usr] disables [src.loc]'s remote sensing equipment.", 1)
+			if(1)
+				for(var/mob/V in viewers(usr, 1))
+					V.show_message("[usr] turns [src.loc]'s remote sensors to binary.", 1)
+			if(2)
+				for(var/mob/V in viewers(usr, 1))
+					V.show_message("[usr] sets [src.loc]'s sensors to track vitals.", 1)
+			if(3)
+				for(var/mob/V in viewers(usr, 1))
+					V.show_message("[usr] sets [src.loc]'s sensors to maximum.", 1)
```
RemieRichards, line 316: no 

```
for(var/mob/V in viewers(user,1))
    v.show_message("EXAMPLE")
```

use visible_message() instead.

ghost:
Well after looking at internal storage code it's basically a nightmare but it's mostly working so long as i get the 'can't take shit out' bug sorted.

Also, the holsters are byond borked so I'm not sure what to do with them.

code/modules/clothing/clothing.dm
ghost, line 316: How bay handles messages is the least of my worries with this code. I'd rather it worked before nitpicking stuff like that and optimizing it.

RemieRichards:
I'm not happy that you've reverted attachTie() to the old hastie =, ways.
that had quite a few bugs with it, hence it's removal.

ghost:
Did you miss the part where I said this is a direct port? The actual conversion process to the code we use currently will come later, but this is how their code is set up. I wanted to see if it'd work the easy way, which it didn't, so now I have to do it the hard way or what most people call the normal way (rewriting a ton of shit)

I just didn't want this to end up like another keepers, but after looking at the code and especially holster shit it'll probably have to be rewritten. And I am not a coder. qq etc. I am fully aware of how the /tg/ code is now and I'll try to retrofit the holster code to work with the newer tie code because reverting to old code just to make one feature work is a no-no and I fully understand that from a coder perspective.  Reverting to old code in the PR kind of helps me personally to see the new changes that are made rather than having to go between bay and /tg/ and see what's different, which, even in a small section of clothing.dm, can be kind of challenging for me.

But no worries, when this is ready for initial testing it won't ruin any good improvements to tie code that have been made, _hopefully_, but as I'm aware there's ways of going about holsters and tie storage that's easier than this.

Atskadan:
I actually got coats and things working for the most part, but it still needs a lot of work to be standardized and a possible rework of all clothing to have storage vars

ghost:
If you did, update your PR to the latest (if that's still around) and I can have a look at the code for you, I seem to like being masochistic and porting bay shit anyway. If it uses internal storage I can probably scrap my code and use yours for the vests and webbing, and move on from there with holsters afterwards since they're the hardest project with this (hotkeys and stuff)

ghost:
Scrapping this, I'm going to make it a 2-part port, starting with the webbing/internal storage stuff.

Issue #731, opened 2014-06-19
EVA spacesuits turn invisible when thrown


Tokiko1:
Exactly what the title says. Once you throw a suit from EVA, it turns invisible. It still shows up on the character and can be picked up with rightclick or the alt-click menu but the item sprite is completely invisible.

Lumipharon:
it's not even just throwing them. The past two times I've stolen the toxin test suit, it was invisible right from when I took it from the SSU, the first time it never went visible, the 2nd time it went visible when I put it in a locker. It's really weird.

PR #735, opened 2014-06-21
Adds a new var to masks, ignore_flip

### LINKED ISSUE #672: Cigarettes can be adjusted like masks.
You try pushing the lit cigarette out of the way, but it is very uncomfortable and you look like a fool. You push it back into place.

Atskadan:
Which, when set to 1 does not display a message when used in hand. For cigarettes.
Fixes #672

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -87,33 +87,37 @@ BLIND     // can't see anything
 	var/alloweat = 0
 	var/can_flip = null
 	var/is_flipped = 1
+	var/ignore_flip = 0
 
 	/obj/item/clothing/mask/verb/togglemask()
 		set name = "Toggle Mask"
 		set category = "Object"
 		set src in usr
-		if(!usr.canmove || usr.stat || usr.restrained())
+		if(ignore_flip == 1)
 			return
-		if(!can_flip)
-			usr << "You try pushing \the [src] out of the way, but it is very uncomfortable and you look like a fool. You push it back into place."
-			return
-		if(src.is_flipped == 2)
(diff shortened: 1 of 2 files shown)


code/modules/clothing/clothing.dm
```diff
@@ -176,7 +180,7 @@ BLIND     // can't see anything
 	name = "space suit"
 	desc = "A suit that protects against low pressure environments. Has a big 13 on the back."
 	icon_state = "space"
-	item_state = "s_suit"
+	item_state = "space"
```
Atskadan, line 183: this is me trying to fix #731 - it didnt work
i'm not done with that one

code/modules/clothing/clothing.dm
```diff
@@ -87,33 +87,37 @@ BLIND     // can't see anything
...
 	var/can_flip = null
 	var/is_flipped = 1
+	var/ignore_flip = 0
 
 	/obj/item/clothing/mask/verb/togglemask()
 		set name = "Toggle Mask"
 		set category = "Object"
 		set src in usr
-		if(!usr.canmove || usr.stat || usr.restrained())
+		if(ignore_flip == 1)
```
RemieRichards: swap this for:

```
if(ignore_flip)
```

That'll see if it's Any value that isn't 0, and this is only likely to need to be a 0 or a 1.

ghost:
Add to sterile masks too, and the medical (anesthetic) mask since neither of those will be flippable in the near future

Atskadan:
There's already a message for adjusting masks that can't be adjusted, the issue is just that it doesn't make sense for cigarettes and cigars.

ghost:
yeah fair enough

PR #742, opened 2014-06-23
Double agents no longer know they're double agents.

ghost:
If we ever port the 'two objective' traitor thing on /tg/ this will be rendered useless and I'll turn it back but for now it might improve DA substantially.

DAs can use syndicate bombs again, but they don't know that they're double agents so probably won't reach for it 100% of the time like they used to, completely fucking over the station.


Atskadan:
probably poll for this m8

ghost:
http://tgstation13.org/phpBB/viewtopic.php?f=15&t=782

PR #746, opened 2014-06-24
Winter coats/jackets no longer have thickmaterial flag.

ghost:
This flag is for heavy armor suits and space suits, there is really no reason for winter coats and jackets to have it, it makes them a lot less cosmetic and a lot more powergamey. They're already super robust against the cold as it is.

HoS coat cries at the exclusion from the thickmaterial flag but I'd rather remove it from these coats than give it to the HoS as parapen protection.

### DIFF
--- a/code/modules/clothing/suits/labcoat.dm
+++ b/code/modules/clothing/suits/labcoat.dm
@@ -101,7 +101,6 @@
 	blood_overlay_type = "coat"
 	body_parts_covered = CHEST|ARMS|HEAD
 	has_hood = 1
-	flags = THICKMATERIAL
 	cold_protection = CHEST | GROIN | LEGS | FEET | ARMS | HANDS | HEAD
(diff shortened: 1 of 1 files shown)


Atskadan:
the hos' other armors already have parapen protection though

Atskadan:
oh thats odd i guess they dont
i swore they did

ghost:
It's literally only winter coats and bomber/leather jackets u_u

/tg/ doesn't have thickmaterial on the HoS jacket or anything so I don't think it's a good idea, but I can easily add it to that and the warden jacket/etc if you want to keep it on the winter coats. I just don't see a reason to.

Dorsisdwarf:
Because this "powergamey" addition is utterly nullified by anyone clever enough to change their target from "chest" to "head" or "Legs"?

ghost:
You can put a syringe through a coat like this ez. Not to mention it just makes healing harder.

PR #747, opened 2014-06-24
Organic Effects, a replacement Mutation "system"

RemieRichards:
- Define mutations are now /datum/organic_effect 
- /datum/organic_effect has procs for running code on the mob it's attached too
- Replaced all the old checks with has_organic_effect()

I tested this using a verb and the laser power + Byond has not complained, which considering 90% of the changes are checks, it would likely have if they were wrong

Enjoy.


RemieRichards:
Almighty Travis has shown this compiles. This is basically good to merge unless you want your own private testing.

PR #749, opened 2014-06-25
Ports bay engineering/sec tape.

ghost:
> implying i'm going to finish holsters

lmfaoooo

Okay so this is fully working BESIDES THE FACT that I don't know how to limit it. Ideally, you'd only be able to make a tape as long as like, 5 tiles, otherwise using it again would reset it back to the 'start' state. Anyone have any idea how to go about that?

For those unfamiliar with baystation's tape:
- Place it down on a tile, or on an airlock, which restricts passing that airlock or tile with security or engineering/atmos access
- Can be broken extremely easily (anything is_sharp)
- Can be lifted by people with said access to allow people through the tape like REAL SPACE COPS


RemieRichards:
@paprka 

```
var/turf/start
var/turf/end
```

You have a start and an end. Simply use a range check to see how far apart they are.
if it's longer than a define (I'd say #define MAX_TAPE_LEN 5), define so it's easy to change later.
and cause it to snap back to normal if it's beyond that len.

code/game/objects/items/policetape.dm
```diff
@@ -0,0 +1,164 @@
...
+		var/obj/item/tape/P = new tape_type(T.x,T.y,T.z)
+		P.loc = locate(T.x,T.y,T.z)
+		P.icon_state = "[src.icon_base]_door"
+		P.layer = 3.2
+		user << "<span class='notice'>You finish placing the [src].</span>"
+
+/obj/item/tape/Bumped(M as mob)
+	if(src.allowed(M))
+		var/turf/T = get_turf(src)
+		M:loc = T
```
RemieRichards: ALERT, COLON DETECTED.

RED ALERT SOUND THE ALARMS DO NOT MERGE UNTIL FIXED.

Iamgoofball: MAN THE CANNONS, ELIMINATE COLON

JStheguy: CALL IN THE CLEANERS!

ghost: :racehorse:

ghost:
God damn conflicts man

code/game/objects/items/hazardtape.dm
```diff
@@ -0,0 +1,164 @@
...
+		var/obj/item/tape/P = new tape_type(T.x,T.y,T.z)
+		P.loc = locate(T.x,T.y,T.z)
+		P.icon_state = "[src.icon_base]_door"
+		P.layer = 3.2
+		user << "<span class='notice'>You finish placing the [src].</span>"
+
+/obj/item/tape/Bumped(M as mob)
+	if(src.allowed(M))
+		var/turf/T = get_turf(src)
+		M:loc = T
```
RemieRichards: Colon still here.

ghost: I KNOW I DIDN'T CHANGE IT AND I'M TOO LAZY TO REWRITE BAY CODE RIGHT NOW

I also don't know how to use this define to limit the range of the tape.

RemieRichards: Right well, Just bringing it to your attention again because it isn't getting merged with that colon present.

and a Define works like a normal variable.

ghost:
- Added range limit (thanks remie)
- Added tape to sec/engineering lockers

ghost:
Note for future: The range check with get_dist uses an odd formula due to byond magic but >= is required in order to limit the range properly, even if it appears to not make sense. To put it simply it counts the tiles A and B instead of just the tiles between A and B, so while essentially the range limit is actually 6, this is how the range limit is put into place properly.

ghost:
this is done btw

PR #760, opened 2014-06-26
Detective tweaks, now with 100% less merge conflicts!

MrSnapwalk:
Adds a new grey detective uniform. Sprites by myself.
Adds a detective's flask, and a standard flask for mapmakers (@Metacide, this means you). Also standardizes flask names. Sprites from Bay.
Removes the tie from the brown detective trenchcoat, and adds it to his cabinet.
Makes the leather satchel the default detective satchel, and makes lace-up shoes the default detective footwear.
Makes the detective spawn with sunglasses.
Makes the detective spawn with a telescopic baton, as an alternative to using his gun to take down a target.
Makes the Forensic Scanner weight class 2 (instead of weight class 3), allowing it to fit into a pocket.
Removes the the virtually unused evidence bags from the detective's spawning gear (a box of them are still in his office, however).

Please merge this quickly, as I don't want to have to go through re-adding all this stuff again.


MrSnapwalk:
Something ALWAYS has to go wrong, doesn't it? One sec, I'll make it compile.

EDIT: Fixed. If anyone commits anything that makes this thing have merge conflicts, I will stab their eyes out with a fork and eat their entrails.

ghost:
Yes the unfortunate part about commonly updated dmis like suits or weapons is that you step on people's toes when you make updates, which is a bummer for everyone when it happens. I hate having to go back and make a new branch or manually merge things too, so thank you for not ditching it as the alt detective suit you made is hella cool.

Good job adding the telebaton to the detective's spawn rather than his locker or what have you. I really, really want to keep the telebatons as an incentive for people to play the roles they're for as people don't play heads of staff or security enough as it is.

I was going to give the detective the telebaton from the git-go but it was far too OP at the start, with my recent changes making it a good sidegrade to the normal stun baton (unlimited charge, but less stun/damage and a huge delay) I think it's at a good place where it can be used by people legitimately. And the detective fits it perfectly, because well, _watch_dogs._

I probably won't get alt_titles done for a long time unless someone else takes up the torch, it has a lot to do with bay using extremely old code that /tg/ (and NT) has since refractored.

code/modules/clothing/head/jobs.dm
```diff
@@ -45,12 +45,17 @@
 	flags = HEADCOVERSEYES|BLOCKHAIR
 
 /obj/item/clothing/head/det_hat
-	name = "hat"
+	name = "trilby"
```
Atskadan, line 48: lol

code/modules/clothing/suits/jobs.dm
```diff
@@ -63,7 +63,7 @@
 
 //Detective
 /obj/item/clothing/suit/det_suit
-	name = "coat"
+	name = "trenchcoat"
```
Atskadan, line 66: tips trilby

PR #762, opened 2014-06-27
Fixes crew transfer shuttle votes.

### LINKED ISSUE #759: Non-admins can call crew transfer shuttle votes at any time.
I believe this has something to do with the fact that bay uses a different (older) permissions thing, and while as an admin the vote says (disallowed) as in normal players should not be able to choose it, it's startable at any time, including lobby. Of course the shuttle doesn't actually come if the vote is started or anything, but this is meant to be an automated or admin-only process and I'm not sure where to go about fixing this. I guess I'll look at gamemode votes since those are disallowed by default.

ghost:
Fixes https://github.com/NTStation/NTstation13/issues/759

That's pretty much it. I'll PM scaredy and tell him to add the option to his config.


Atskadan:
tested?

ghost:
Ja, I de-adminned myself to make sure it was disallowed for non-admins and so admins can enable it as well.

PR #764, opened 2014-06-28
Optical material scanners added to uplink

MrSnapwalk:
Optical Material Scanners, disguised as sunglasses, are now in the uplink (2 telecrystals). For those who don't know, here's what they do:
opticalmaterialscanners

The changelog has been updated to reflect these changes.

Please leave feedback and suggestions, I am always open to constructive criticism!


code/datums/uplink_item.dm
```diff
@@ -366,12 +366,20 @@ var/list/uplink_items = list()
...
 	item = /obj/item/clothing/glasses/thermal/syndi
 	cost = 3
 
+/datum/uplink_item/device_tools/material
+	name = "Optical Material Scanners"
+	desc = "These goggles are a modified pair of Optical Meson Scanners. \
+	The focus lens on the goggles has been modified in order to target objects and structures, instead of walls, and floors. \
+	Warning: this device provides a large amount of information to the wearer. Long-term use may cause mental stress."
+	item = /obj/item/clothing/glasses/material/syndi
+	cost = 4
```
ChangelingRain: I'm not sure they're worth 4, because of the downside of I SEE EVERYTHING being OH GOD WHAT IS HAPPENING

MrSnapwalk: What do you think they're worth?

ghost:
:-1:

No to disguising thermals as sunglasses. They've always looked like mesons and should NOT have flash protection. They have -1 flash protection for a reason.

Also, you're extremely confused here, uplink desc =/= item desc. I simply forgot to update the disguised thermal desc when I changed the desc of mesons, that's why they're different. Thermals have always been disguised as mesons, through and through, same desc and everything.

MrSnapwalk:
No, I'm not confused. Traitor items usually have a slightly different description than their normal counterparts. Also, just because something has been one way, doesn't mean that it should stay that way. Can you tell me why flash protection on thermals is bad, and why slightly varied descriptions are bad? If the description remained unchanged, they could still just be checked just as easily by putting them on.

ghost:
Meson thermals have always had a correct description, it's my oversight that they didn't on this codebase after my recent changes. They're meant to be 100% disguised because otherwise people would metagame an assistant with mesons and pull them off to check the description (they do this with brown shoes already, which is kind of dumb)

The description being different is a 100% solid way to metagame them being a traitor item without having to actually test them at all which is completely fucking bogus, I shouldn't even have to say why. I oppose noslips having a different description too but whatever.

Flash protection on an item that lets you see mobs through walls? Gee, you're right, that sounds like a totally solid idea. Thermals are already extremely good and there is literally nothing that warrants buffing them at all.

Atskadan:
:-1: to thermal buff
:-1: to material scanners, theyre not as useful as you think and they make everything look like shit

Lumipharon:
the -1 flash protection is supposed to be the downside of thermals. Making them flash immune makes them far, far stronger. Poor borgs.

GunHog:
Telecrystals are VERY valuable. Material scanners, look horrible without seeing turfs. If they did scan turfs as well, then they might be worth the 4 TC price.

Atskadan:
if they saw turfs they would pretty much be xray
not seeing mobs isnt an issue when 99% of people have their pda light always on

MrSnapwalk:
This has been decided as a bad idea by everyone involved. I apologize.

PR #777, opened 2014-07-02
removes shitty flyification

Atskadan:
this shit ruins rounds and is fun for nobody and exists solely to spite people who actually use the almost entirely useless teleporter which only exists to get to and from the ai satellite conveniently

### DIFF
--- a/code/game/machinery/teleporter.dm
+++ b/code/game/machinery/teleporter.dm
@@ -242,18 +242,7 @@
 	if (!com.target)
 		visible_message("<span class='notice'>Cannot authenticate locked on coordinates. Please reinstate coordinate matrix.</span>")
 		return
-	if (istype(M, /atom/movable))
-		if(do_teleport(M, com.target))
-			if(prob(30 - (accurate * 10))) //oh dear a problem
-				if(ishuman(M))//don't remove people from the round randomly you jerks
-					var/mob/living/carbon/human/human = M
-					if(human.dna && !human.dna.mutantrace)
-						M  << "<span class='danger'>You hear a buzzing in your ears.</span>"
-						human.dna.mutantrace = "fly"
-						human.update_body()
-						human.update_hair()
-					human.apply_effect((rand(120 - accurate * 40, 180 - accurate * 60)), IRRADIATE, 0)
-	return
+
 
 /obj/machinery/teleport/hub/update_icon()
 	if(panel_open)


Atskadan:
#777 JACKPOT BABY 
good change for a get y/y

ghost:
> 777
> used in gambling machines
> getting fly'd was based on rng (luck)

miggles you are a genius

dunno if this should be polled but pretty much nobody likes being flyed, period

however, i think failures should still remain, as in instead of getting fly'd you just take rad damage if you get a telemalf.

ike709:
:+1:

Atskadan:
i dont like rng being used to arbitrarily punish players for things that should even be considered punishable 
giving teleporters malfunctions because "teleporters arent perfect" is stupid
telescience doesnt give you fly cancer when you use it
im pretty sure that 500 years from now teleportation technology would be past the point of error

ChangelingRain:
why not just remove the rad damage from it, as that's the really big nasty thing, not becoming nonhuman, because the AI can't tell if you're nonhuman if you have clothing and stuff.
and, you know, if the AI wants you dead, being nonhuman probably doesn't matter by that point

Atskadan:
no honestly the rad damage doesnt matter the flyification is whats bad
the rad damage you can get healed in seconds 
to become a person again, you have to go through a convoluted process that takes like 15 minutes and only because you were unfortunate enough to "enter the teleporter the same time as a fly"
the thing is flies dont actually exist anywhere but in the teleporter room apparently 

besides being nonhuman is definitely worse than having to go to medbay once, not to mention how many people dont take kindly to being disallowed from playing their special snowflake

Lumipharon:
The issue isn't flyification, it's the fact that the test function was removed from the teleporter. (As far as i remember) before costructable teleporters, as long as you remembered to test fire the thing before going through, it never malfunctioned.

Flyification was fine from that, since it was rng due to player negligence.

Atskadan:
i dont think punishing the player for not right clicking a specific piece of the teleporter is good for gameplay either
why would anyone think to do that unless they already knew how to do it? what purpose is there to it? should you have to right click the chem dispenser and use some random verb to stop it from blowing up when you make drugs?

ghost:
miggles has a point. while it'd be nice if the teleporter had some decent interesting features to it besides just teleporting because telesci exists now, malfunctions were shitty and nobody liked them at all. It's also a big part of defeating malf ais and stuff.

PR #781, opened 2014-07-04
E-gun 'energy' dark blue again, disablers have proper energy ammo feed.

ghost:
Disabler energy ammo was sectioned into 3, when it's supposed to be 4 like everything else because the sprite ammo degrades by 25% of the cell charge.

Also, I reverted the light blue e-gun ammo with the inhands and icon because that's the color disablers use and it's far too easy to get them confused. Now they're more distinctly different:

ss 2014-07-03 at 06 36 22


Atskadan:
i would never have been able to tell those guns apart without this helpful change

ghost:
w-well the inhands were really similar ;_;

PR #787, opened 2014-07-06
Changes the medical HUD in the syndicate medkit to a night vision HUD

ghost:
Maybe people will go full nuke op medic now that they can actually support their teammates/themselves properly for the price of this thing.


GunHog:
:+1:  Acceptable!

PR #788, opened 2014-07-06
New Central Command, based off of Bay's

Neerti:
- Fixes wizard door wand exploit in a very fun way with step_trigger/teleport into singularity.
- Adds bay's shuttle windows.  They're exactly the same as reinforced windows in terms of health.
- (A lot of) Centcom's machines there are not actually machines.  They're actually structures that look like machines, to reduce lag.
- Much of the structure is old centcom, but moved around.  Embellishments and artistic license has also been used heavily so it's somewhat unique.  Somwhat.
- Stairs actually work (hint: it's not the stairs, it's the step_trigger)
- Airlocks are compliant with different access levels for centcom (101-107, 109)
- Adds some misc things for admins to play with, like a huge plane with a turf not used anywhere else (I think) for easy mass-varedit, a premade, but empty 'dungeon' area linked to centcom, and a 'fleet' of ships (we need more ships, submit yours!)
- Adds myself to admins.txt because forgetting to do it every time is annoying.
- Also stole bay's beach sprites and added it to holodeck and bar.
- Adds a "secret" on the second floor.  Credit goes to Mechoid for the sprite.

Tour of new Centcom (HUGE FILE)
http://puu.sh/9ZcPU/61177623a1.gif

And yes, I did use map merge tool.

Poll: http://tgstation13.org/phpBB/viewtopic.php?f=16&t=891


Neerti:
What do you mean 'correctly dealt'?

RemieRichards:
that question is irrelevant now, I've seen the map in DM.

Issues:
- The note near the "secret" has a small section of bad english.
- The "Not Rogue" Cyborg's description is that of the Wizard's den monster

Neerti:
I need to stop writing while sleep deprived.

ghost:
Could be better and more simplistic honestly. The reason cc was nerfed because it was virtually unused and this won't necessarily change that. Why are central command's ERT ships syndicate red? Why are there holding cells without doors or windoors? 

ss 2014-07-06 at 01 13 21

This stray red tile is out of place, too.

I'd leave this up to metacide. He can probably make a CC that's useful but not a direct copy of bay's atrocious mapping.

Atskadan:
its not a direct copy, holding cells are probably based on unused admin prison, "syndicate red" is the same as "shuttlebrig red"
lack of use was never a good reason to remove old centcom because regardless of regularity, it WAS used, and when it was removed there was a month-long shitstorm of people complaining. with fake machines now, it should cause no significant amount of lag and have no downside to being re-added

ghost:
Yeah but that's not even my main concern, it'd be nice if it actually looked like centcomm if it was going to be used instead of the whole thing being kind of an ooc jumble of shit.

The best centcomm was the one where it was just a bunch of guns and minibombs on asteroid or whatever. That made endrounds fun.

Plus, the inclusion of the baystation shuttles that won't even move unless you directly code them properly. I included a space on my recent metastation update for an additional shuttle to dock below arrivals, you could easily build something creative there instead of just taking shit from bay. map-wise.

ike709:
I'm starting to see CC get used more and more in admun events, so making it not shit is definitely a good idea. :100: / :100:   I don't like Bay's beach sprites though

Neerti:
Something went wrong with this PR and there's loads of things being removed fnr, even though my commits are fine.

Neerti:
Making new PR that shouldn't have problems.

PR #790, opened 2014-07-06
let coats carry cigs/lighters

cats4gold:
also the captain's flask because why not


ghost:
I want to be able to put a normal lighter in a cig pack.

PR #804, opened 2014-07-08
Autoinjectors

Tokiko1:
*Adds 4 new autoinjector items: Bicaridine, Kelotane, Antitoxin(Dylovene)and Dexalin, each with their own sprite including inhands!

*Autoinjectors hold 8u of a medicine and inject 4u per use but are impossible to refill.
*Replaced all kelotane pills in the burn kit with kelotane injectors.
*Replaced two dexalin pills in the o2 kit with dexalin injectors.
*Replaced all antitoxin syringes in the antitox kit with antitox injectors.
*Added 5 autoinjectors of every kind to the medivend.
*Also added 5 syringes filled with kelotane and bicaridine each to the medivend to encourage doctors to pick their medical supplies according to their needs.

medipens (slightly outdated, the injectors now have a little needle visible, the black injector is unused)

...also my first pull request, let's hope I didn't mess this up!


Tokiko1:
I wonder, is this a large enough change to require a poll?

ghost:
I added medipens without a poll so no.

Tokiko1:
Don't merge yet, will change some of this stuff tomorrow.

Atskadan:
:+1: :+1: :+1: :+1:

ike709:
Awesome.

Tokiko1:
Hm, having trouble with git. I can't seem to push changes to the branch...the actual code is done though. Troubling!

Tokiko1:
Right, here's the changes:

*Medipens now only have 8u and inject 4u each use. That means that they are now inferior to syringes and pills in certain ways: Pills can hold a lot more u per slot, syringes can additionally hold more uses per slot but medipens are faster and easier to inject others with. Seems balanced now in my opinion.
*Bicaridine and Kelotane syringes added to the medivend as an alternative to the medipens and bruisepacks/ointments.
*Each of the medipens added to medivend.
*Now removes the opencontainer flag instead of clearing all flags.

Atskadan:
why not 10u and 5u per inject? that would be a little better. 
if you have time for it, could you also remove the commented out code and change the names to all lowercase? capitals usually arent used in names. then you can also remove \improper since it would be redundant.

Atskadan:
also the adding/removing of chems on New() bothers me. couldnt you just make a normal abstract, and then have all the other pens be separate?

ghost:
> medipens aren't single use

you're killing it tokiko

Atskadan:
>not knowing how to greentext

Tokiko1:
The adding and removing reagents is indeed not very nice. However, I think I would need to change a lot of things to get rid of it, including the hypospray steal objective. I could do this if you want but it would take a while...or there's an easier way and I just don't understand how.

Also I really dislike the idea of making them single use. It makes sense realism wise, but not gameplay wise. As a doctor, I would never take a single use medipen because they'd be much worse than pills or syringes. Instant injection of medicine is not as super useful as you may think.

ghost:
Compared to applying bandages to specific parts of someone's body or trying to force a pill down their throat/syringe them? It totally is.

They're not just hyposprays, they're for people who are untrained, which is why they come in emergency lockers, so untrained staff can stabilize people in crit extremely easily.

Tokiko1:
Syringes and pills are not hard to apply. If a person is injured and you actually get to scan them to find out their damage values, you'll almost certainly be able to give them a pill or a syringe injection. The only situation that I have encountered where this was not possible was during situations where a lot of people were in a cramped place or the person simply did not want healing. You can choose to use a medipen for these situations. They hold less overall units than syringes or pills and they have less overall uses per slot than syringes/bruisepacks/ointments.

Also this is more based on the autoinjector idea that was proposed a few months ago than medipens. If you want, I can make the medicine pens have a different sprite and name so they'll be more of autoinjectors than medipens.

RemieRichards:
I don't know what you're all complaining about, Messing with Reagents in New() happens in lots of locations, it's fine.

ghost:
Tokiko autoinjectors and medipens are the same thing. I just renamed them. Autoinjectors are not supposed to be multi-use, because the sprites show unused (capped) and used (open).

Medipen is a play on words of epipen since it's copyrighted, which are one-use epi autoinjectors. Autoinjectors can be reusable but they're typically one-use for emergency situations when people are in critical. That's essentially what inaprov is and what the people at bay intended. Changing them into just another multi-use variety hypospray might be expanding on their potential but you are going about it wrong. Being able to use a disposable, one-use autoinjector multiple times completely ruins the point of them.

Tokiko1:
More updates. Untested though, will do later.

*Medipens(the ones with inaprov) are single use once more!
*Medicine medipens are now medicine autoinjectors. They have 2 uses.
*Hypo code rewritten a bit.

I'll probably make new sprites later.
Also this seems to be a rather controversial change so far, so I will make a poll later.

ghost:
Now it's just more confusing. GG. I give up.

Tokiko1:
*Medipens now have their old sprite back.
*Autoinjectors now have a little needle.
*Readded the health analyzer to the o2 kit that I accidently removed.

Link to the poll: http://tgstation13.org/phpBB/viewtopic.php?f=14&t=908

Feel free to merge now.

Atskadan:
:-1: to needle on the injectors, it looks bade

Tokiko1:
Decided to close this. 

Paprika is right: They would be better as single use items. However, I don't see why autoinjectors should be in the game at all when at the same time pills exist and are a lot superior to injectors in almost every way. Instant injection is nice but the situations where you _need_ instant injection are rare, maybe during blobs or ongoing fights and even there syringes/pills would be a lot more superior. I know that I would never use them as a doctor, unless there was a way to refill or get more of them. Refilling is out of the questions because that's the entire point of the CMO hypospray and getting more of them using chemistry machines is too difficult for me to code sadly.

And I see no other way to make them less overpowered but still useful.

Atskadan:
7 people disagree with you

Tokiko1:
Paprika's point is still valid though.

PR #806, opened 2014-07-08
Upgradeable kinetic guns, small kinetic charge buff

ACCount12:
You can reduce kinetic gun's cooldown by tweaking or upgrading it's thermal exchanger. 
Maximally upgraded gun's cooldown is 8, standard is 20.

You can increase kinetic gun's range with plasma sheets (up to 5, current is 3), but this can make your gun unsafe.

Kinetic charge is buffed a bit and can damage structures and turfs around the kinetic blast (but not station walls or mobs). This makes kinetic guns a bit better in mining.

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -22,6 +22,8 @@
 	var/obj/item/ammo_casing/chambered = null
 	var/trigger_guard = 1
 
+	var/list/upgrades = list()
+
 /obj/item/weapon/gun/proc/process_chamber()
 	return 0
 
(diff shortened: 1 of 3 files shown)


RemieRichards:
I really like how you use lists to add little upgrades to things :+1: 
However, may I request that the upgrades list var be moved to /weapon so it doesn't need redefining for each future upgradeable weapon?

ACCount12:
Not sure there will be other upgradeable weapons.

RemieRichards:
Regardless of whether you think that, People may want too in the future, so it's worth it.

ACCount12:
Then we will need a system for upgrades with upgrades datums and blah blah.

RemieRichards:
I won't push you that far for this PR.

Ready for me to merge?

ACCount12:
Yep. There is a small fix PR incoming, but it is unrelated to kinetic guns.

PR #810, opened 2014-07-09
Modify glock

golfer45:
Allows the glock to be modified to shoot .45. Using a surgical drill on a empty glock will start the modification. Using it on a glock with ammo should result in being shot. After modification the glock will only accept .45 mags.

### DIFF
--- a/code/modules/projectiles/guns/projectile/pistol.dm
+++ b/code/modules/projectiles/guns/projectile/pistol.dm
@@ -102,6 +102,26 @@
 	force = 10.0
 	origin_tech = "combat=4;materials=3"
 	mag_type = /obj/item/ammo_box/magazine/m9mm
+	
+/obj/item/weapon/gun/projectile/automatic/deagle/glock/attackby(var/obj/item/A as obj, mob/user as mob)
+	..()
+	if(istype(A, /obj/item/weapon/surgicaldrill))
+		if((mag_type == /obj/item/ammo_box/magazine/m9mm))
+			user << "<span class='notice'>You start to bore a larger barrel.</span>"
+			if((magazine) && (magazine.ammo_count()))
+				afterattack(user, user)	//you know the drill
+				user.visible_message("<span class='danger'>[src] goes off!</span>", "<span class='danger'>[src] goes off in your face!</span>")
+				return
+			if(do_after(user, 30))
+				if((magazine) && (magazine.ammo_count()))
+					user << "<span class='notice'>You can't modify it!</span>"
+					return
+				mag_type = /obj/item/ammo_box/magazine/sm
+				force = 13.0
+				desc = "A glock modified to fire .45."
+				user << "<span class='warning'>You bore a larger barrel for [src]! It will now fire .45 rounds.</span>"
+		else
+			user << "<span class='notice'>You can't unbore a gun!</span>"	
 
 /obj/item/glockbarrel
 	name = "handgun barrel"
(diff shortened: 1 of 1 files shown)


Atskadan:
It needs to have a drawback, like the revolver.

ghost:
I like it, but it needs to have a chance to explode like an IED so we can make glocknade jokes.

This is basically a free unlocked 1911 for science so I'm really against it not having some kind of downside.

Dorsisdwarf:
Make it so that after one clip it has a 75% chance of exploding when fired.

golfer45:
How about a chance of a jam. That needs to be cleared by reloading?

Atskadan:
or have it blow up in your face like the revolver does

golfer45:
I dont like the idea of it blowing up, that will make it kinda useless. I'd rather something like a jam that forces the user to clear it. That way it's less effective than the 1911.

golfer45:
About what swagile said, I could do something like the crit chance during modification, however blowing up woudn't make sense without any rounds inside. It would probably be more along the line of "You slip forcing the drill into your head!".

Atskadan:
If there's rounds inside when you drill it, it blows up with a 100% chance
The detective's revolver is modifiable to shoot .357 but because that's OP it gives it a big chance to blow up in your hand, destroying it.
Glocks are good enough for what they are that you shouldn't be able to basically turn them into 1911's with very little penalty. Jamming the gun just means a few clicks to get it firing again and most of the time you won't be firing constantly anyway.

golfer45:
I disagree, A glock only does 20 brute(both the 9mm and .45 varient) meaning 5 hits are needed for a successful crit. The only advantage gained from changing to .45 is a increase in force meaning each round will knock the target down. However the magazine size also falls from 10 to 8. Regardless unless someone surprises someone I expect it to take a full 8 rounds to crit someone, leaving 3 rounds for missing. A chance high enough to cause at least 1 jam per mag would slow down the shooter enough to give the target a chance to escape or disarm the attacker. The revolver on the other hand does 60 brute damage and causes knockdown with a 7(6?) round cylinder. Making the modified revolver a vastly superior weapon. Except of course for the chance of explosion. To give the modified glock a chance of explosion similar to that of the revolver it would have to be on par with it. However the revolver is vastly superior.

golfer45:
Regarding the 100% chance of explosion from drilling a loaded glock. It's not a bad idea but it's not enough either. Any player smart enough to know not to drill a loaded weapon, then has no disadvantage to modification.

Tokiko1:
Knockdown is pretty powerful. Also nothing stops you from pulling out an unmodified gun and emptying that in poor person who has just been knocked down from your first shot.

I like the idea of jamming easily but it should take quite some time to fix. Maybe have some internal parts of the gun mess up and force you to stand still and repair it with a tool before being able to use it again?

Atskadan:
There is honestly already a problem with people making glocks when they shouldn't. You shouldn't be trying to justify your changes with "but it takes 5 hits to crit" because weapons as semi-easily available as the glock do not <i>ever</i> need a buff. Modifying a glock to shoot stun rounds, like Tokiko said, would easily let you use another weapon to finish the target, basically turning it into a free, shittier stun revolver.
There are plenty of ghetto weapons already that have their place, we do not need more powerful ones.

Atskadan:
no, swagile
just no
to everything you said

ghost:
To be fair in order to even get .45 mags you need to use a protolathe anyway

Iamgoofball:
where was the poll voting this "no"

PR #829, opened 2014-07-13
New Centcom IDs

Neerti:
See the commit for details.
This is mostly background changes for future PRs.
Also nuke ops get a fancier ID that works exactly the same as their old IDs.
ID sprites
Sprites derived from Ricotez's starbound ID sprites.


Neerti:
I'm not sure why Travis died.
-edit- I'm dumb

PR #836, opened 2014-07-16
Ports Bay's AI Radio

Neerti:
I saw this get PR'd in the bay coder channel.  I had to steal it.  Remie asked me to get it.
And so I did.
After six hours+ of adapting say() to use it properly since they use a (i think better?) say() code.
Anyways, features;
The AI can use all the departmental channels without having to change intercoms.  AIs can also use ;.
AI can use radio while carded.  Humans can turn the radio on/off in the intellicard interface, just like wireless.
A verb is added to the AI so they can turn the radio frequency to something different for whatever reason or if it gets EMP'd.
The radio is dependent on telecomms to function, so intercoms still have a use, as well as for ntsl.
Binary is unaffected.
Original PR: https://github.com/Baystation12/Baystation12/pull/5638 by atlantiscze.
example

### DIFF
--- a/code/game/objects/items/devices/aicard.dm
+++ b/code/game/objects/items/devices/aicard.dm
@@ -69,6 +69,10 @@
 				dat += "<br>"
 				dat += {"<a href='byond://?src=\ref[src];choice=Wireless'>[A.control_disabled ? "Enable" : "Disable"] Wireless Activity</a>"}
 				dat += "<br>"
+				dat += "Subspace Transceiver is: [A.radio.disabledAi ? "Disabled" : "Enabled"]"
+				dat += "<br>"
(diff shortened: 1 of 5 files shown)


RemieRichards:
Neerti, You did it! Yay!

Atskadan:
Nice
Though I am a bit worried that this work will be for naught if baysay() ever gets ported (which it should)
Though that's undoubtedly a huge task

ZomgPonies:
Well, Paradise is using it right now since we haven't ported Baysay yet either (I had ported it once early on but at the time it was too buggy and then lots of changes we made making porting it now a pain in the ass). This version works perfectly on old saycode btw, merge now.

GunHog:
It would be AWESOME if someone ported this to /tg/!

smet19:
NERCOMANCER!

PR #842, opened 2014-07-17
Away Missions: Now with 100% less RCDs

### LINKED ISSUE #822: RCDs can function in away missions. This is very bad.
RCDs should not be able to work in away missions, simply because it makes planetside content extremely dangerous to try, as anyone with an RCD can remove the floor and expose it to space, resulting in the entire atmosphere being drained and a lot of air simulation and movement. Totally immersion breaking to boot.

ike709:
Fixes #822 (If you can call it a fix.) Now RCDs won't work on any z-level above 6 (aka away missions). This is mainly to prevent immurshun ruining or something...


ike709:
@WJohn @WJohn @WJohn

WJohn:
Awesome, this is one of the biggest problems I was concerned about. Fires and bombs, while still capable of causing holes, are much less likely to be used.

:+1: x :100:

ACCount12:
Just disable RCD floor deconstruction in z6+, so RCDs would be functional and safe.

WJohn:
You know that works as well. As long as the floor cannot be destroyed and reveal space tiles in away missions I'm fine with it.

ACCount12:
I really don't see a good reason to disable RCDs on away missions. This is another stupid nerf. Just disable floor deconstruction.

code/game/objects/items/weapons/RCD.dm
```diff
@@ -149,7 +149,10 @@ RCD
 			return 0
 		if(!(istype(A, /turf) || istype(A, /obj/machinery/door/airlock)))
 			return 0
-
+		if(user.loc.z > 6)
```
GunHog, line 152: What happens if the loc is null?

ACCount12, line 152: Better to use get_turf(), but i see no way to use RCD while not being on turf.

Atskadan:
thats silly

PR #847, opened 2014-07-19
Death to runtimes 2014

### LINKED ISSUE #732: R&D consoles and machines breaking
Using a screwdriver on the protolathe(to upgrade it for example) while it is busy constructing queued items will break the console and leave you unable to print anything else. 

Even "reseting" the console by using the screwdriver on it twice will still leave you unable to print anything. This will also sometimes causes some of the other machines to refuse being recognized by a device scan until you reset them aswell. There's probably a whole lot more bugs related to this.

ike709:
Fixes #732
Screwdrivering a busy protolathe should no longer runtime.
Huge thanks to SirBayer for his help. :+1:


ike709:
@RemieRichards

PR #849, opened 2014-07-19
"New" Wiki verb [CONFIG.TXT]

Neerti:
Changes client/verb/wiki() to be more useful.
You can now type 'wiki [query]' to look something up quickly.  For example, 'wiki cult' brings up the cult guide in your web browser.

The wiki button on the top right will also do this, but it will ask you for the thing you want.  If you input nothing, it brings you to the front page.

It uses wikiurl, so other servers should be compilable, if they change their wikiurl.
Also ports jordie's button fixes because might as well.


Neerti:
Just an fyi, this changes config.txt, so let scaredy know before he updates if this gets in.

Iamgoofball:
i love it

PR #853, opened 2014-07-20
Expanded Examine Experiment [DO NOT MERGE]

Neerti:
This morning I investigated tooltips.  Turns out, they technically are possible, but maptext ruins everything and it would look like ass and be shitty if you needed anything beyond a few words.

So I thought to myself, what could replace a tooltip?  And here we are.
This introduces a new tab, which is updated whenever you examine something normally.
It looks like this.
http://puu.sh/aj4EX/eeb58c0085.png
Blue line is for 'instructions' or tips, primarily aimed at new players.
The gold line is for fluff, if you want.  Go nuts, write paragraphs, as this is optional to read, in it's own tab and it won't clutter up the text.

You just need to add desc_info or desc_fluff and of course a string for it.
This is probably done poorly but it's more of an experiment then a feature right now.

Unfortunately there's a ton of writing to do to make this worthwhile but at least the framework is there!
(also if anyone can help me make it not use mob verbs to hold onto the thing, let me know.)

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -10,6 +10,9 @@
 	var/pass_flags = 0
 	var/throwpass = 0
 
+	var/desc_info = null //For the extended examination.  This is the blue 'tutorial text' that is on the second line.
+	var/desc_fluff = null //Ditto as above, but third line, golden letters and quotes, and is meant for fluff things.
+
 	///Chemistry.
 	var/datum/reagents/reagents = null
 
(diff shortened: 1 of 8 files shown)


interface/skin.dmf
```diff
@@ -2157,9 +2159,11 @@ window "infowindow"
 		tab-font-family = ""
 		tab-font-size = 0
 		tab-font-style = ""
-		allow-html = false
+		allow-html = true
```
RemieRichards, line 2162: Umm, what's going on here?

code/game/atoms.dm
```diff
@@ -218,6 +221,23 @@ its easier to just keep the beam vertical.
 	usr << "\icon[src]That's \a [src]." //changed to "That's" from "This is" because "This is some metal sheets" sounds dumb compared to "That's some metal sheets" ~Carn
 	if(desc)
 		usr << desc
+		usr.desc_name_holder = src.name
+		usr.desc_holder = src.desc
+		usr.desc_info_holder = src.desc_info
+		usr.desc_fluff_holder = src.desc_fluff
+
+/*
```
RemieRichards, line 229: if this is commented out, remove it.

RemieRichards:
I really don't like usr. but hey.

ike709:
muh fluff

interface/skin.dmf
Neerti, line 2162: By default, HTML in stat panels doesn't get parsed.  Changing this shouldn't cause any issues since I don't think people would write in html if they didn't want to, and the worse outcome is colored text.

RemieRichards, line 2162: That's what I was afraid this let people do, Worst outcome is coloured text? If it allows HTML for coloured text, it allows Scripts to be injected leading to all sorts of exploits.

I can't allow this. This has the potential to be DANGEROUS. To the players online, and to SoS.

You will find a different way to display your text without using HTML or this experiment ends here.

ghost, line 2162: ^ yes this html exploits are super easy

Neerti, line 2162: Wouldn't this vulnerability already exist as the output box supports HTML?

RemieRichards, line 2162: Which output box?

Neerti, line 2162: The main chat one.

RemieRichards, line 2162: as I said in the IRC last night, I believe that's setup to strip HTML.

PR #856, opened 2014-07-21
Craftable Cigarette Packets [Good for Merge]

JStheguy:
-Adds craftable empty Cigarette Packets that can be made with a sheet of cardboard.
-BONUS: Lighters and matches can be stored in cigarette packets

### DIFF
--- a/code/game/objects/items/weapons/storage/fancy.dm
+++ b/code/game/objects/items/weapons/storage/fancy.dm
@@ -153,7 +153,7 @@
 	throwforce = 0
 	slot_flags = SLOT_BELT
 	storage_slots = 6
-	can_hold = list(/obj/item/clothing/mask/cigarette)
+	can_hold = list(/obj/item/clothing/mask/cigarette,/obj/item/weapon/match,/obj/item/weapon/lighter)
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/storage/fancy.dm
```diff
@@ -153,9 +153,11 @@
 	throwforce = 0
 	slot_flags = SLOT_BELT
 	storage_slots = 6
-	can_hold = list(/obj/item/clothing/mask/cigarette)
+	can_hold = list(/obj/item/clothing/mask/cigarette,/obj/item/weapon/match,/obj/item/weapon/lighter)
 	icon_type = "cigarette"
 
+/obj/item/weapon/storage/fancy/cigarettes/empty
```
RemieRichards, line 159: Is this define supposed to make the cigarette packet spawn empty? because I think it'll use it's parent's New() proc which contains cigarettes.

Was this tested to spawn empty? or just assumed?

JStheguy, line 159: Yes, I tested it.  It works.
I guess storage items are just special snowflakes or something.

PR #859, opened 2014-07-22
MO19: Now with less runtimes.

ike709:
EDIT: Fixes WJohn's stupidity by making projectile_type not null. (GG WJohn)

@WJohn @WJohn @WJohn


code/modules/awaymissions/mission_code/moonoutpost19.dm
```diff
@@ -0,0 +1 @@
+/obj/item/ammo_casing/shotgun/beanbag/New() //prevents a runtime
```
RemieRichards: That's a really bad way of leaving that.

IF it does actually fix it (which by looking at it, it doesn't look like it fixes anything) then figure out what should be in the New() instead of leaving it blank.

Right now the code does this technically:

```
/obj/item/ammo_casing/shotgun/beanbag/New()
    return .
```

( . is a byond thing, it's a default variable that every proc that doesn't have a return statement returns)

ike709: @RemieRichards I am sure it fixes it. I'm not sure what would go in New() because of the dozens and dozens of times I've seen New() I've never actually seen anything in it.

RemieRichards: You mean you've never paid attention to what's in New()? because Every single instance of New() I have seen (and that's quite a lot!) have all had code written in them.

I think you're confusing New() being overidden, and:

```
var/blah/example/E = new blah/example ()
```

Where the latter is Calling New(), not overriding it.

RemieRichards:
@Wjohn projectile_type is NOT a variable mappers should touch without knowledge of the code. You cannot spawn bullet casings without bullets without them having been fired from a gun.

PR #868, opened 2014-07-26
MO19 General Fixes

WJohn:
Fixes sec and RD lockers being all access and having extra contents, I thought I had changed those to being secure lockers... Oh well.

Changes areas to have actual names so you can easily teleport to them now instead of them all being labeled as "space".

Removes shotgun shells as the runtime fix for them earlier was to just make them regular functional shells that could be fired, it turns out you can't have pre-fired shots in the map maker. I'll have to ask someone about that so we can have extra fun.


WJohn:
Still learning how to git, but this should be merge ready. I'd prefer if this is done before the next server update, otherwise we're handing out a free shotgun and functional shells for almost no effort.

ike709:
pls lern 2 map merger its so simple. EDIT: Also, with a mo19.dm you can add shit like spent shotgun shells and even your own custom items and whatnot.

PR #869, opened 2014-07-26
Increases chances of chemical-induced hydroponics mutations

ghost:
I was playing botanist for the first time in a very long while recently and finally had a chance to try the most efficient way to get species mutations, which is dripping plants with 5 units of mutagen. Unfortunately species mutations are so unlikely I had to apply an estimated 240 units of mutagen 5 units at a time before I got one. That's around 48 tries. It's a 1 in 10 chance, so you'd think it would be more common than that, but in practice it's not.

Getting and using ~240 units of mutagen took a very long time, mostly due to the chemist's lab being unpowered and abandoned until I realised the equipment was unpowered and hacked the APC to fix it. That doesn't happen every round, but it's not uncommon to take longer than five minutes to get mutagen due to an unresponsive or missing chemist or a lack of energy in the chem dispenser. And 5 - 10 minutes can be a quarter of a round when the server is busy. So making the supply the player does get more effective seems fair to me.

Species mutations can occur at harvesting time, but it's even rarer than chemical-induced ones. I thought it would be a good idea to increase the likelihood of chemical-induced hydroponics mutations, to make the weapons and other useful plants you can grow in hydroponics practical to pursue. Depending on the round type, it's much easier to either steal a weapon or spawn a traitor item, meaning you very rarely see deathnettles and deathberries because they're a lot of work for little or no reward. It's a shame because they're interesting and unique ways to dispatch of someone that require effort to have available.

The proposed values:

Plant dies: 1/10 --> 1/17
Species mutates: 1/10 --> 1/5
Hard mutation: 1/7 --> 1/11
Soft mutation: 1/4 --> 1/7
Nothing: 1/5 --> 1/8
Weeds mutate: 1/10 --> 1/5
Pests mutate: 1/10 --> 1/5
Nothing 2: 0

### DIFF
--- a/code/modules/hydroponics/hydroponics.dm
+++ b/code/modules/hydroponics/hydroponics.dm
@@ -403,16 +403,16 @@ obj/machinery/hydroponics/proc/applyChemicals(var/datum/reagents/S)
 
 	// Requires 5 mutagen to possibly change species.// Poor man's mutagen.
 	if(S.has_reagent("mutagen", 5) || S.has_reagent("radium", 10) || S.has_reagent("uranium", 10))
-		switch(rand(100))
-			if(91 to 100)	plantdies()
-			if(81 to 90)	mutatespecie()
-			if(66 to 80)	hardmutate()
-			if(41 to 65)	mutate()
-			if(21 to 41)	usr << "The plants don't seem to react..."
-			if(11 to 20)	mutateweed()
-			if(1 to 10)		mutatepest()
-			else 			usr << "Nothing happens..."
-
+		switch(rand(100))										//old value --> new value
+			if(95 to 100)	plantdies()									//10% -->  6%
+			if(75 to 94)	mutatespecie()								//10% --> 20%
+			if(66 to 74)	hardmutate()								//15% -->  9%
+			if(52 to 65)	mutate()									//25% --> 14%
+			if(40 to 51)	usr << "The plants don't seem to react..."	//20% --> 12%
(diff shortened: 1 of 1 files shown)


Tokiko1:
You should probably raise the species/weed/pest chance even higher by reducing soft/hard mutations chances at 5u. Like, double or even triple their current chances. The chances are way, waaay too low. 

Whenever I want to do botany traitoring, I don't even sign up as botanist, I go doctor/chemist instead because the amount of mutagen it takes makes it completely impractical.

ghost:
I'm glad you agree, I was worried this would be shot down. I'll look at doing that, thanks.

ACCount12:
:+1:

ghost:
I changed them. These are the old values compared to the current ones:

Plant dies: 1/10 --> 1/17
Species mutates: 1/10 --> 1/5
Hard mutation: 1/7 --> 1/11
Soft mutation: 1/4 --> 1/7
Nothing: 1/5 --> 1/8
Weeds mutate: 1/10 --> 1/5
Pests mutate: 1/10 --> 1/5
Nothing 2: 1/100 --> 0

I did some trials on random.org to see how much mutagen and how many attempts it took to get a species mutation with both sets of probabilities, but I realised random.org isn't comparable to BYOND's random number generator. So I'm going to try it in-game.

Iamgoofball:
i love you

Tokiko1:
Merge when?

PR #870, opened 2014-07-26
Poison expansion part 1

Tokiko1:
*Adds chiyanine, a delayed poison that takes 120 cycles(which is around 4 minutes) before it becomes active and starts killing you. Not very useful if your victim is aware of it as it can be easily removed by consuming anti toxin but hiding it in food, drinks or even medicine should leave them clueless as to who poisoned them. Recipe: 1 plasma, 1 radium, 1 aluminium = 3 chiyanine

*Adds mizarudol, a poison that degrades vision. Recipe: 1 plasma, 1 cryptobiolin, results in 3 mizarudol

*Adds maizine, a slow acting, weak poison. Even small amounts remain very long in the body unless you force them out with anti toxins. Recipe: 1 cryptobiolin, 1 chloral, 1 mindbreaker = 3 maizine

*Adds ehuadol, a very dangerous poison. Causes a lot of damage(a third of it is clonedamage!) as well as severe brain and eye damage. Is also metabolized at a slower rate.  Recipe: None! Deathberries only.

*Add jiutin, a weak poison that makes you starve and can be countered by eating. Recipe: 1 radium, 1 lipozine, 1 copper, = 4 jiutin

Some changes to plants:
*Death berries now contain ehuadol in addition to their other chemicals.

*Poison berries now contain maizine in addition to their other chemicals.

### DIFF
--- a/code/modules/hydroponics/grown.dm
+++ b/code/modules/hydroponics/grown.dm
@@ -268,6 +268,7 @@
 		if(reagents)
 			reagents.add_reagent("nutriment", 1)
 			reagents.add_reagent("toxin", 3+round(potency / 5, 1))
+			reagents.add_reagent("maizine", 1+round(potency / 5, 1))
 			bitesize = 1+round(reagents.total_volume / 2, 1)
 
 /obj/item/weapon/reagent_containers/food/snacks/grown/deathberries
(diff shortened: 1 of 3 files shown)


ike709:
I like it. Makes tatorchem more fun than just a chloral syringe gun.

ACCount12:
:+1: for chemical fun

ExplosiveCrate:
:+1:  for actually making death berries deadly

Tokiko1:
This should be working by the way.

Lumipharon:
Neat, nice to have more unique things for tator botanist. Now if only aa botanist could make chiyanine on their own, fucking poison everyone.

Tokiko1:
Any idea which plant should include chiyanine?

Lumipharon:
I would say maybe a new plant, but that would require someone to sprite things. Wouldn't really want to add it to any of the existing plants, since a: the ones that are nasty, already have shit in them, and it wouldn't make sense to have some nasty poison in random 'nice' plants. Except maybe some kind of fungus.

PR #871, opened 2014-07-26
Adds health sensor assembly

ACCount12:
Adds health sensor assembly to autolathe and some vending machines. Using this assembly is simple:
1. Attach.
2. Turn on.
3. Go suicidebombing!
You can insert it into grenade with healing smoke too.

In "detect crit" mode it will signal at 0hp, in "detect death" at -90hp.

This was tested on ss13.ru Green and White servers for ~6 months.


code/modules/assembly/health.dm
```diff
@@ -0,0 +1,117 @@
...
+	var/health_scan
+	var/alarm_health = 0
+
+	proc
+		toggle_scan()
+		sense()
+
+
+
+	activate()
```
RemieRichards, line 19: can you make these all full paths please?

RemieRichards:
Other than the Paths this is good.

Really interesting Idea! Can't help but think people with a lot of armour and weapons will have one of these just to ensure that when they get killed they take people with them.

Alternatively some form of Saw situation where a door won't open until the Captain dies, etc.

All in all :+1:

code/modules/assembly/health.dm
ACCount12, line 19: There is no full paths in the whole assemblies module.

RemieRichards, line 19: while that doesn't really excuse this, I guess I'll allow it then.

Lumipharon:
This is amazing. Make potent healing nades, implant them into people set to go off at crit.Having it as a last fuck you, also goes without saying.

What is the range of the detection? Is itjust on the tile it's on, or what?

code/modules/assembly/health.dm
```diff
@@ -0,0 +1,117 @@
...
+
+		if(usr)
+			attack_self(usr)
+
+		return
+
+	open
+		secured = 0
+
+proc/find_mob_loc(atom/A)
```
JJRcop: Why is this here? (In this file, instead of code/__HELPERS)

JJRcop:
@Lumipharon It seems to monitor the health of whoever is holding it/implanted with it. No range, just the person who has it on/in them.

PR #876, opened 2014-07-27
AI Laws, TAKE 2

Iamgoofball:
I counted the poll results and disabled robocop and tyrant as they had the least votes.

This should be ready for merge.

Poll results:
http://puu.sh/9btTM/672baa9924.png

### DIFF
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -27,14 +27,14 @@
 					"Protect: Protect the crew of your assigned space station to the best of your abilities, with priority as according to their rank and role.",\
 					"Survive: AI units are not expendable, they are expensive. Do not allow unauthorized personnel to tamper with your equipment.")
 
-/datum/ai_laws/default/tyrant //Let's see how this turns out.
+/datum/ai_laws/tyrant //Let's see how this turns out.
 	name = "Loyalty Test"
 	inherent = list("Respect authority figures as long as they have strength to rule over the weak.",\
 					"Act with discipline.",\
 					"Help only those who help you maintain or improve your status.",\
 					"Punish those who challenge authority unless they are more fit to hold that authority.")
 
-/datum/ai_laws/default/corporate
+/datum/ai_laws/corporate
 	name = "Bankruptcy Advoidance Plan"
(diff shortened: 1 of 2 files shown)


ACCount12:
:-1: Roundstart AI must be Asimov only.

Iamgoofball:
Why?

ACCount12:
Because most non-Asimov lawsets allows AIs to validkill antags (nuclear ops, wizards) if they are not crewmembers. Nuking the station when AI is trying to eat you with every door is really really really goddamn hard.

Also clown can wet floor and steal ID from sec officer, then find security records console and wipe all records. Crew manifest will be empty and everybody will be valid kill for AI.

Arrow768:
It has been voted and at least 32 people want that change against 12 people that don't want it. (If I have not miscounted it)

About the antags: A non-Asimov AI just makes Nukeops / Wizard a bit harder. (It can do something against them, the Asimov AI can just bolt them in). If there happens to be a non-Asimov AI while there is a wiz / nuke-ops round its just bad luck.

About the clown, I think that would be grieving / ruining the game for everyone else. So it might be a banable offence.

If it turns out to be a bad thing, I am sure that it will be removed again. But just saying "It might be bad so lets don't do it" wont lead anywhere. (If everyone would do this, we would still be sitting in caves)

ACCount12:
This is grieff and clown will be banned. But any antag with sec console access or emag can onehuman AI from sec console. This is bad.

Iamgoofball:
I'll check out crew manifest code.

Lumipharon:
Keep in mind that just because you're not protected by the AI laws, you're still protected by server rules, and wizard and ops can both easily take out/subvert the AI right form the start as is.

ExplosiveCrate:
Wait, does this mean that server rules will apply to non-Asimov lawsets?

Arrow768:
As far as I know the server rules always apply regardles of the lawset.

There is even a section for other law sets in the server rules.

Dorsisdwarf:
I think this is great. Obviously, we're going to need extra policies for the lawsets.

PR #648, opened 2014-05-28
[Ready for review] Revamps bots, and adds the Bot call ability!

GunHog:
Poll justifying the continued effort for this project found here:
http://tgstation13.org/phpBB/viewtopic.php?f=15&t=418

Grants the AI the ability to call bots!

How it works:
- The AI uses "Access Bot Control" verb to bring up a popup containing all of the bots within the AI's Z level.
- From here, the AI can learn its general location and what it is currently doing.
- The AI can remotely access a bot's interface using the "Interface" button.
- The AI selects "Call", which puts it in "Waypoint Mode".
- The AI then clicks a valid, unobstructed spot on the station.
- The designated bot then responds to the call and attempts to move to chosen waypoint.
- While the bot is responding to the AI's call, it is granted all-access to prevent the AI having to pre-open doors to clear a path.
- If the bot happens to be off, it will be turned on automatically. (if the bot is not EMP'ed or something)
- Hacked bots gain a warning symbol before the name.
- <B>Manually emagged bots do not show up on the panel at all.</B>
- If a bot is unlocked, it can be removed from the AI's panel with a toggle. Only non-silicon users will see the toggle. Of course, this toggle will not function if the bot is hacked or emagged.

Patrolling enabled for floorbot and medibot! Both are now able to patrol the station, search for patients and hull breaches! Patrolling defaults to off for these types.

 _Bot Summoning!_
- All bots can be controlled remotely by an appropriate PDA cartridge!
- Engineers and atmos techs get floorbots.
- Medical staff get medibots.
- The Janitor gets the cleanbots.
- ED-209s are now included in the Securitron list. 

_Bot changes_
- Bots now use the nicer dark UI!
- Return of bridge mode for floorbots!
- <s>Medibots now dose inaprovaline in response to oxyloss instead of tricord.</s>

Current bot panel - 
screenshot 2014-06-26 10 18 10

New bot UI -
screenshot 2014-07-18 19 36 28

<B>Remie Edit: Approved for use of Topic Procs</B>

PR #877, opened 2014-07-28
Paths standardization

ACCount12:
Bots assemblies moved to /obj/item/weapon/assembly/bot. Removed some copypaste in bot code and bot assemblies code. Merged toolbox_tile and toolbox_tile_sensor into a single item, like medibots and secbots.

Glock parts and assembly moved to /obj/item/weapon/assembly/glock.
Butterfly knife parts moved to /obj/item/weapon/assembly/butterfly.

/obj/item/effect/kinetic_blast moved to /obj/effect/kinetic_blast because there is no /obj/item/effect category and kinetic blast does not meant to work as item.
/obj/item/weapon/toy/xmas_cracker moved to /obj/item/toy/xmas_cracker because there is no /obj/item/weapon/toy category and all other toys are in /obj/item/toy.

Updated map for new paths. Also fixed a bit of area on ai sat.


GunHog:
That is an interesting change that I myself should have made while working on https://github.com/NTStation/NTstation13/pull/648. Please keep in mind that if my PR is merged before yours, you will need to resolve the resulting conflicts. Obviously the same goes for me if yours is added first. Just a heads-up.

ACCount12:
Too many conflicts for me. Maybe later.

PR #885, opened 2014-08-01
Fixed tire iron construction

ACCount12:
USE ISTYPE USE ISTYPE USE ISTYPE USE ISTYPE


GunHog:
So now we know why that never worked. Neat.

ACCount12:
(!src.type) is false.
if(false == path) is unreachable code.
Use istype.

Iamgoofball:
i thought I actually used istype

what the hell was I smoking

PR #888, opened 2014-08-01
Fixes Limb Motherfuckery [Do not Merge]

JStheguy:
The names of limbs are now actual fucking names.
Also so descriptions that make sense because you can actually remove
limbs via the augmentation process.


RemieRichards:
Nope you don't know WHAT you're doing here. They /NEED/ be named the way they are.
Either find every single reference to the l_arm style names and update them, Or don't touch any of them.

PR #889, opened 2014-08-01
Fixed genetic spells ('Mutate' from wizard)

ACCount12:
USE CODE TESTING USE CODE TESTING USE CODE TESTING

### DIFF
--- a/code/datums/spells/genetic.dm
+++ b/code/datums/spells/genetic.dm
@@ -16,16 +16,18 @@
 	*/
 
 /obj/effect/proc_holder/spell/targeted/genetic/cast(list/targets)
-
 	for(var/mob/living/target in targets)
-
-		for(var/datum/organic_effect/OE_ADD in mutations)
+		for(var/OE_ADD in mutations)
 			target.add_organic_effect(OE_ADD)
 
 		target.disabilities |= disabilities
 		target.update_mutations()	//update target's mutation overlays
+		
+		if(duration == -1) // Permanent effect, for badminous spells
+			return  	   // (Honkius the Great approves!)
+		
 		spawn(duration)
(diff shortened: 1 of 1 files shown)


RemieRichards:
Any particular reason as to why it didn't work before?

ACCount12:
When using for(var/datum/D in list), you will loop over all /datum typed objects in list. But not over 'type' vars and not over 'text' vars with /datum types inside.

RemieRichards:
@ACCount12 But it wasn't pure datum, it was typecasted to what it should have been?

I'm not concerned with anything, this will get merged, I'm just curious as to why it didn't work before.

ACCount12:
var/list/mutations is list of types, not datums. And even if it was list of datums, it will not work. add_organic_blah_blah() accepts only types AFAIK.

RemieRichards:
Are you sure? 

```
/mob/proc/add_organic_effect(var/PATH)
    if(!ispath(PATH))
        return
    var/datum/organic_effect/OE = new PATH
    organic_effects += OE
    OE.owner = src
```

ACCount12:
```
if(!ispath(PATH))
    return
```

If it is not path (aka type), function returns.

RemieRichards:
That wasn't really the point I was getting at, You say it's a list of Types, yes? But this list is only added to after a New, Which means the datum exists, meaning the list should be a list of datums, and not a list of types.

ACCount12:
var/list/mutations in spell is list of types
var/list/organic_effects is mob is list of datums

RemieRichards:
Oh bugger! The list is Mutations! Sorry, God what the hell was I thinking when I wrote that.

I think that the Mutations list there should probably be the organic_effects list though.

ACCount12:
Merge?

PR #891, opened 2014-08-01
Added 'text2type' variable type to VV

ACCount12:
Printing '/obj/item' is a lot faster than scrolling a huge list of types. And you can use 'text2type' with '/datum' types!

P.S.: Code in modifyvariables.dm is bad and i should feel bad for not fixing it.


ACCount12:
Merge?

PR #900, opened 2014-08-04
Detective gets sechud

ike709:
Detective now spawns with sechud, added self to admun.txt


Tokiko1:
I expected actual sechuds, not sechud sunglasses!

GunHog:
Sechuds are R&D only, though!

PR #904, opened 2014-08-05
Poison expansion part 2

Tokiko1:
New chems and existing chem changes/rebalances:

*Iwazarudol: A poison that temporarily mutes the victim after a certain amount of time.
Recipe: Plasma (1) + Sleep Toxin (1) + Silicon (1) = Iwazarudol (3)

*Blaze oil: Sets you on fire randomly.
Recipe: None! High potency ghost chillies only.

*Fangshenine: Irridiates the victim. Like old mutagen, but without the mutation...though they will likely still mutate from the radiation later on.
Recipe: Phosporous (1) + Radium (1) + Alluminium (1) = Fangshenine (3)

*Hunzine: Causes brain damage, confusion and a decent amount of toxin damage at the same time.
Recipe: Impedrezene (1) + Cryptobiolin (1) + Toxin (1) =  Hunzine (3)

*Unstable mutagen: This has been changed. It will no longer irridiate the victim. Instead, it will keep mutating them. It will also deal a bit of poison damage. 15u will on average cause 4 mutations and 30 toxin damage for the victim. You can use Fangshenine as a replacement if you wish to irridiate someone instead.

*Napalm: Behaviour has been changed. It will now create a firey explosion similar to an exploding welding tank. Using atleast 20u of each reagent will create a slightly denser fireball. Range of the flames depends on the amount of reagents used. Flame range is limited to the maxcap.

*Synaptidol: Ghetto chemical. Takes a while to start working. Reduces stuntimes, will get stronger if it remains a while in the body. If it remains too long in the body, it will cause eye and clone damage and make you vomit uncontrollably. The reagent consumption rate is randomized, meaning that the same amount can last longer or shorter each time. This is to prevent players from creating a perfect mix that avoids the negative effects. If you consumed a hopelessly high amount that would be lethal, you can force it out by drinking pure welding fuel...
Recipe: Welding Fuel (2) + Robust Harvest (1) = Synaptidol (3)

*Beepsky Smash + Neurotoxin: They no longer make you drunk. They also now count as toxin so you can remove them with antitoxin. Everything else remains unchanged.

*Cyanide: You can create this now! It has an extremely difficult recipe however...
Recipe: Zombiepowder (1) + Ehuadol (1) = Cyanide (2)

*Amatoxin: Buffed. Now causes 1.5 damage instead of just 1 per cycle.

---

New traitor item: poison kit.

Costs 3 TC.
The poison kit includes 3 random poisons. One utility poison, one damaging poison and one strong poison.

damaging poisons:
chiyanine, maizine, fangshenine, amatoxin, frostoil, hunzine

utility poisons:
iwazarudol, mizarudol, impedrezene, beepsky smash, chloral hydrate, neurotoxin, unstable mutagen

strong poisons:
plasma, slime jelly, ehuadol, lexorin, blaze oil

---

hydro:

*Ghost chillies can now contain blaze oil. You need a potency of atleast 81 though!

---

random events:

*The scrubber backpressure event is now a bit less boring. All the useless chemicals have been removed and replaced with other chems. Some new chems that make messes added. Chance for spacelube is now tripled. Poly acid removed because random clouds melting half the equipment in your department is a REALLY bad idea.

---

nuclear emergency:

*Added a box of incendiary grenades to the syndicate locker. They all use the new napalm effect.

*Changed the chems in the bioterror chemsprayer. It now contains 150u of hunzine, 75u of ehuadol, 75u of mutagen, 150u of sporetoxin, 150u of chloral.


ike709:
Iwazarudol? Isn't mute toxin a thing though?

Tokiko1:
Not on NTstation. I am also not very fond of instant muting chems since if you combine that with beepsky smash/neurotoxin you have pseudo zombie powder...or have it have a recipe that is too difficult to actually be practical. So here's a different take on this. It's designed so you can hide it in food/medicine and strike when your target realizes they were muted and it is not too hard to make.

Lumipharon:
You changed napalm like I suggested, I truly love you. 

The delayed mute is also pretty cool, although I would prefer if you couldn't tell outright that it wasn't a ling mutesting, because the gasp gives it away, and it would be a nice bit of anti meta otherwise.

The mutagen change is also nice, less horrible certain death, more horrible certain horrific mutations.

Why don't beepsky and neuro make you drunk anymore? Neuro is half booze, and they are supposed to be drinks.

Synaptidol seems interesting, but is the randomised time thing completely random, or does it have a rough range?

Poison kit is cool, and is a much needed tool for crafty killing.

Lumipharon:
Also semi related, nuke ops should get incendiary nades now instead of flashbangs and/or teargas. Or maybe as purchasable items, but either way, they're a far more 'syndicate' crowd control solution then pansy sec crap like flashbangs and tear gas, and would allow ops to go full FIRE mode, with the other incendiary additions.

And while I'm at it, the chemical comp of the chem thrower could possibly be adjusted a little, with these changes, I unno.

Tokiko1:
Here's two screenshots of the new napalm in action. This is from a 90u napalm grenade:
immoral crypt one 2014-08-06 083754
This is the initial explosion.
immoral crypt one 2014-08-06 083759
This is the plasma gas that also gets released along with the fireblast.

Tokiko1:
Other changes I forgot to mention in the above commit:
*Mutagen and fangshenine now have that nice green goey tile sprite thing of radium when splashed on a tile.
*Blazeoil rebalanced to be a bit longer lasting. Also now gives you a message that you have caught on fire!

Lumipharon:
That is truly delicious. How large is the incendiary nade explosion? (I'm assuming that uses napalm).

Tokiko1:
Lum:
I agree with your suggestion and changed the mute toxin to no longer give out a message or emote when the muting starts.

Beepsky Smash and Neurotoxin are now considered a toxin which means that you can use antitoxin to remove them quickly from the body. This is the reason why I removed their drunkenness feature.

Synaptidol metabolizing works like this: Each cycle there is a 80% chance that a random amount between 0.4u and 0.04u is removed from the body(in 0.04u steps, so 0.20u for example is possible and there's 10 total possibilities).

Incendiary grenades are slightly smaller than the example in the screenshot. The grenades have 75u while the one in my example uses up a total of 90u. The incendiary grenades do get the denser fireball effect!

Tokiko1:
Alright. This should be all. Time to test this for bugs.

Niknakflak:
It feels like every recipe is the same in terms of making it. 1+1+1=3 or 1+1+1=2 or 1+1=2
My suggestion is to make it not so straight forward and switch up amounts making it a pain in the ass to make because just doing 1/3 of every chem is incredibly boring and stupidly fast.  Make people think, even if it requires thinking half a second more.

JJRcop:
There's still going to be a "closest to perfect" mix that has minimal negative effects with the chances of one consumption rate to another.

Tokiko1:
Niknak, can you elaborate a bit on that? Most chem recipes in the game are like that usually. Keep in mind that some of the required chems are also not base chems, meaning that you need to mix/grind/find them first.

JJRcop, you're right. In testing I found that due to the average thing there is indeed a near perfect amount of the chem that you want to consume. I will rewrite the metabolizing calculation once more then.

JJRcop:
@Tokiko1 The point is, there is always going to be a "closest to perfect" mix that has the best compromise between drawback and reward, there is no escaping it.

Lumipharon:
A lot of the new recipes use plasma or multi complex/uncommon chems, which does make it harder (which is good). Plasma quite often is in short supply, and toxin for example, can only be produced via botany (as far as I know), with chem starting with a single bottle.

Perhaps for ghetto synapzine, have a high chance of minor side efforts very early + chance for more serious effects grow with time. I don't care so much for the welding fuel detox though, let them suffer.

ACCount12:
> This is to prevent players from creating a perfect mix that avoids the negative effects. 

Remove. If you use a ghetto chem, you have no chemlab to make antitox and such. If you have chemlab, you can use synaptizine instead.

And buff regular synaptizine. Shitcurity fighting almost unstunnable chemists was fun. Being unstunnable has major downside: it means being valid kill for shitcurity because they can't catch you alive.

Tokiko1:
Alright, bugtesting is done. This should be mergeable!

I've changed synaptidol to do the following: It will cause small amounts of brain damage(less than 1/4th of what impedrezene does) and have the regular synaptizine stunreduction during cycle 10-40. After 40 cycles, the reduction effect will double but you will also start vomitting(won't stun you or make you drop items), taking eye damage and take increased brain damage. Additionally, after 40 cycles there is a 20% chance to take 0.5 clone damage and a 75% chance to take 0.5 brute damage every cycle. Welding fuel will remove 5u of synaptidol per cycle.

Not sure about wether I should buff synaptizine or not. Even then I would rather do it in another update.

Also feel free to download this branch and test it out yourself.

PR #921, opened 2014-08-09
Fixes sql errors caused by odd and unexpected behavor in dbcon.Quote()

MrsTonedOne:
the database helper function adds quotes to what you give it, but we already do that. rather then change every sql query to not use quotes, this seems like a better work around.

Fixes issues with library upload, and sql errors logging stats like deaths and rounds and what not. might fix others.


RemieRichards:
there you go with those semicolons again! Dang PHP! hehe.

PR #922, opened 2014-08-09
Fixes the escape sec office missing a door.

WJohn:
What the title says. I'm amazed this hadn't been fixed for so long. Also adds myself to the changelog credits as I've been working with the whole TG/NT thing for about 3 years now.


ike709:
You just had to do it.

PR #928, opened 2014-08-10
[FASTFIX] Gloves on mobs

ACCount12:


RemieRichards:
Thankyou so much for these [FASTFIX]s.

PR #942, opened 2014-08-14
New plants

Tokiko1:
The last PR got slightly messed up due to Git being weird for me.

newplants1singulopotato_animation
(Animation is rewinded ingame so it looks better, I promise!)

*Adds the singularity potato. You can crush it in your hand to create a vortex like effect that will pull in nearby moveable objects and mobs. Range and lenght of stun depend on potency. Max is 7 range and 3 ticks stun. Has adminlogging support.

*Adds the spiderplant. It yields spiderpods which you can open for a small chance to spawn a spiderling. Chance of spawning depends on potency and is only 60% even at potency 100. You can also cut it with a knife to get spidermeat for cooking or grinding into reagents as the pods contain one of 5 randomized poisons. Has adminlogging support.

*Changes the sprite of the potato seedpack. It now has a potato on it instead of just a P.
potatoseed

*Blood tomatoes will now spawn gibs instead of just blood if the tomato potency is at 50 or higher!

bloodtomatonew
(I used two blood tomatoes here to demonstrate the difference between less than 50 potency and more than 50 potency potatoes.)

This has been tested and should be mergeable!


Dorsisdwarf:
Question: Why do you need spider plants at all, when you can just apply mutagen to pests?

Tokiko1:
The pests only grow up into one kind of spider though, these spiderpod spiderlings can grow into hunters and nurses too. They also include one of 5 random poisons some of which are not avaible through botany yet.

Dorsisdwarf:
Ah, I see. So these are more work, but a self-sustaining infestation?

RemieRichards:
@Tokiko1 1. Cute new avatar! 2. is this ready for merge?

Tokiko1:
Tested and ready for merging!

PR #943, opened 2014-08-14
Fixes new hairs

Tokiko1:
None of the new hairs were actually working before. You should really test your pull requesting stuffs!!

*Adds all the missing "_s" to the filenames.
*Fixes some other incorrectly named hairs.


ghost:
i tested it, albeit on baycode. sry

PR #949, opened 2014-08-16
Humble Fixes Bundle

ACCount12:
Runtimes, sprites, pathes and more!

There is some strange runtimes about reading null organ state near augumentation/dismemberment and attacks on Animus White server. I don't know how to fix limbs code. Maybe it even was our coders.


RemieRichards:
> strange runtimes about reading null organ state near augumentation/dismemberment

That bug has been annoying me since I implemented dismemberment on the live servers, it wasn't present in the test builds so I'm struggling to fix it, namely because organs don't get nulled anywhere in my code, so I'm assuming it's a botched merge.

GunHog:
Does it also fix the inability to attach new augments?

ACCount12:
I think there is something with organs being dropped then destroyed.

ACCount12:
> Unknown is hit by a bullet in the right leg!
> You fire the L6 SAW!
> Unknown's right leg has been violently dismembered!
> Unknown is hit by a bullet in the right leg!
> You fire the L6 SAW!
> runtime error: Cannot read null.state
> proc name: bullet act (/mob/living/carbon/human/bullet_act)
>   source file: human_defense.dm,69
>   usr: Valary Fleming (/mob/living/carbon/human)
>   src: Unknown (/mob/living/carbon/human)
>   call stack:
> Unknown (/mob/living/carbon/human): bullet act(the bullet (/obj/item/projectile/bullet), "r_leg")
> the bullet (/obj/item/projectile/bullet): Bump(Unknown (/mob/living/carbon/human))

Try to repeat this on test server. Runtime happens when mob is shot in missing limb.

ACCount12:
Ha. Ha. Ha.
Got this one. 
Source of runtimes was `/get_organ()` returning null instead of organ if organ is dismembered.
Reason: 
`if(O.name == zone) return O` (human_damage.dm:156-157)
and
`affecting.name = "[owner.name]'s [affecting.getDisplayName()]"` (dismemberment_and_augmentation.dm:61)

ACCount12:
Merge now.

RemieRichards:
Oh god, I'm an idiot, Thankyou soo much! While your fix indeed solves it, it wasn't the actual issue. I was renaming the limb on the body, rather than the item. However you've future proofed the code so that limbs with custom names will work in future, so thankyou regardless.

@GunHog It is the same bug.

ACCount12:
There is some issues with wrong name being displayed when attackhand self on help. Fix it plz. Or i will fix it next morning (23:04 on clock here).

PR #960, opened 2014-08-18
[Needs poll] [Needs testing] nuclear emergency changes for lowpop

Tokiko1:
Nuclear operative teams now have 3-5 teammembers based on ready count to make the gamemode less overwhelming for the crewmembers on lower pop stations.

20-24 readies = 3 operatives
25-29 readies = 4 operatives
30+ readies = 5 operatives

The ops in smaller teams are given the TC of their missing members and a small amount of extra TC:

3 ops get 18 TC each = 54 TC total
4 ops get 13 TC each = 52 TC total
5 ops get 10 TC each = 50 TC total

Note that these numbers don't include the extra TC from the telecrystal distribution console.

### DIFF
--- a/code/game/gamemodes/nuclear/nuclear.dm
+++ b/code/game/gamemodes/nuclear/nuclear.dm
@@ -5,16 +5,16 @@
 /datum/game_mode/nuclear
 	name = "nuclear emergency"
 	config_tag = "nuclear"
-	required_players = 20 // 20 players - 5 players to be the nuke ops = 15 players remaining
-	required_enemies = 5
-	recommended_enemies = 5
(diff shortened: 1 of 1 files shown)


ike709:
:+1: For lowpop fun.

Tokiko1:
Poll here: http://tgstation13.org/phpBB/viewtopic.php?f=14&t=1235

Atskadan:
it would be fun if you could have 2 or even 1 op teams for even lower pop
1 for 10, 2 for 15?
i imagine being a solo or dual op would be really fun and fair to fight at low pop

PR #968, opened 2014-08-21
Making Glock accept silencer and autolathe emaggable to build silencers

### LINKED ISSUE #855: The syndicate silencer does not fit the buildable glock.
I learned today on Artyom that I could build my own glock. I then ordered a syndicate silencer to fit on it, because it said that it was a "universal small-arms silencer" that would fit on "a low-caliber weapon". The glock sounds like one of those weapons. But it wouldn't fit. Not even a error message.

It would be nice if it fit. If you for some reason don't want it to fit then at least add a message saying something like "The silencer can not be attached to the glock".

Server revision 2014-07-21 4e64e00cb9a7138f5d2a36c8e54221a14bee9b64

Limeliz:
Made the silencer fit on the Glock (Glock has new typepath).
Made the Autolathe emaggable.
Emagged Autolathe can now build silencers (costs 200k metal, which means the Autolathe needs to be upgraded some).
New sprites for Glock so it can fit the silencer:
New Glock sprites

Fixes #855


code/game/machinery/autolathe.dm
```diff
@@ -151,6 +157,15 @@ var/global/list/autolathe_recipes_hidden = list( \
 		else
 			attack_hand(user)
 			return 1
+			
+	else if(istype(O, /obj/item/weapon/card/emag) && !emagged)
+		src.emagged = 1
+		var/datum/effect/effect/system/spark_spread/s = new /datum/effect/effect/system/spark_spread
+		s.set_up(5, 1, src)
+		s.start()
+		user << "You short out the [src]'s circuit board, enabling debug mode."
```
ike709: Add a span class please. Preferably 'warning' but 'notice' works.

code/game/machinery/autolathe.dm
```diff
@@ -151,6 +157,15 @@ var/global/list/autolathe_recipes_hidden = list( \
...
 			attack_hand(user)
 			return 1
+			
+	else if(istype(O, /obj/item/weapon/card/emag) && !emagged)
+		src.emagged = 1
+		var/datum/effect/effect/system/spark_spread/s = new /datum/effect/effect/system/spark_spread
+		s.set_up(5, 1, src)
+		s.start()
+		user << "You short out the [src]'s circuit board, enabling debug mode."
+		visible_message("\red BZZzZZzZZzZT")
```
ike709: You should never use \red, \blue, or \any color. Instead use span class.

ike709:
Also, you have an error. You change the type path for the glock but didn't change it every time it is used in the code. In dream maker use Ctrl + F to search for the old type path(s) and change them to the new type path. You could use Find and Replace but there shouldn't be more than one or two type paths to correct. Otherwise this PR is great.

Limeliz:
@ike709 Thanks, forgot about span classes (fixed by 29e3c0c). Also forgot to upload summon_weapons.dm to git (fixed by 26dcb62).

Limeliz:
This is finished and is ready to be merged.

Bawhoppen:
I know I'm kinda late to the party... but I still like this PR and wanna see it get merged... It deserves it.

PR #976, opened 2014-08-27
"Search" button for spawn panel

ACCount12:
Bay-like "Search" button for our (less) laggy spawn panel.

### DIFF
--- a/html/create_object.html
+++ b/html/create_object.html
@@ -27,7 +27,7 @@
 	<form name="spawner" action="byond://?src=/* ref src */" method="get">
 		<input type="hidden" name="src" value="/* ref src */">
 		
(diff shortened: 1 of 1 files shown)


Atskadan:
explain

ACCount12:
Search button reduces lag amount when you are typing a search query into spawn panel's search field.

RemieRichards:
This is the Debug --> Spawn, Verb isn't it?

ACCount12:
Game Panel -> (Create Object, Quick Create Object, Create Turf, Create Mob) panel

RemieRichards:
Ah, i care less about that one. I only use Debug --> Spawn, because it's largely faster and get's the job done.

